PR #13, opened 2020-06-01
bugfix where opening expense in Fyle routes to an invalid URL

ashwin1111:


apps/fyle/utils.py
```diff
@@ -33,6 +36,39 @@ def get_employee_profile(self):
...
+        """
+        Get cluster domain name from fyle
+        """
+        access_token = self.connection.access_token
+        api_headers = {
+            'content-type': 'application/json',
+            'Authorization': 'Bearer {0}'.format(access_token)
+        }
+        body = {}
+        api_url = settings.FYLE_BASE_URL + '/oauth/cluster/'
```
Shwetabhk, line 49: use string formatting
`'{0}/oauth/cluster/'.format(settings.FYLE_BASE_URL)`

PR #18, opened 2020-09-29
allow only one state to be selected in expense states

Sravanksk:


apps/fyle/models.py
```diff
@@ -142,10 +142,8 @@ class ExpenseGroupSettings(models.Model):
         help_text='list of fields ccc expenses grouped by'
     )
 
-    expense_states = ArrayField(
-        base_field=models.CharField(max_length=100), default=get_default_expense_states,
-        help_text='list of states to fetch expenses'
-    )
+    expense_states = models.CharField(max_length=100, default=get_default_expense_states,
```
Shwetabhk, line 145: rename the field to expense_state
help text is wrong

apps/fyle/models.py
```diff
@@ -124,7 +124,7 @@ def get_default_expense_group_fields():
 
 
 def get_default_expense_states():
```
Shwetabhk, line 126: call it `get_default_expense_state():`

apps/fyle/models.py
Shwetabhk: When you rename the field it will ask for a default value while running `makemigrations`
put `PAYMENT_PROCESSING` there

Sravanksk:
it will not ask for default value while running migrations because we are already assigning a default value to `expense_state` in the model itself 
`default=get_default_expense_state` 

`get_default_expense_state` is a function that is already defined above 
`
def get_default_expense_state():`
      `return 'PAYMENT_PROCESSING'`

apps/fyle/migrations/0004_expensegroupsettings.py
```diff
@@ -20,7 +20,7 @@ class Migration(migrations.Migration):
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
```
Sravanksk, line 20: made changes to this file because - 
this migration file is throwing an error since this is dependent of function `get_default_expense_states` 
But we renamed this function to `get_default_expense_state`

Shwetabhk (approved):

PR #21, opened 2020-10-05
Slack logs for Quickbooks API

Shwetabhk:


fyle_qbo_api/settings.py
```diff
@@ -106,23 +105,44 @@
...
+        'request_logs': {
+            'class': 'logging.StreamHandler',
+            'stream': sys.stdout,
+            'formatter': 'requests'
         },
     },
     'loggers': {
-        'django.request': {
-            'handlers': ['console'],
+        'errors': {
```
jayachandhar, line 136: use django or add a wildcard entry like '*' as logger

fyle_qbo_api/settings.py
```diff
@@ -106,23 +105,44 @@
...
+            'handlers': ['debug_logs'],
             'level': 'ERROR',
-            'propagate': False,
+            'propagate': False
         },
-    },
+        'django': {
```
jayachandhar, line 141: use django.request for request logging

apps/quickbooks_online/tasks.py
```diff
@@ -24,7 +24,7 @@
     JournalEntry, JournalEntryLineitem
 from .utils import QBOConnector
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger('errors')
```
jayachandhar, line 27: use module name

apps/fyle/tasks.py
```diff
@@ -13,7 +13,7 @@
 from .utils import FyleConnector
 from .serializers import ExpenseGroupSerializer
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger('errors')
```
jayachandhar, line 16: use module name

fyle_qbo_api/settings.py
```diff
@@ -106,23 +105,44 @@
...
     'version': 1,
-    'disable_existing_loggers': False,
+    'disable_existing_loggers': True,
+    'formatters': {
+        'verbose': {
+            'format': '{levelname} %s {asctime} {module} {message}' % SERVICE_NAME,
+            'style': '{',
+        },
+        'requests': {
+            'format': 'request {levelname} %s {asctime} {message}' % SERVICE_NAME,
```
jayachandhar, line 119: use different format for request logs 
p0: log the following: api path, status code ,response size, time taken, some id to identify the user
p1: log the following: user agent, referrer

jayachandhar (changes requested):

PR #34, opened 2021-01-25
added filter for task

NileshPant1999:


apps/tasks/views.py
```diff
@@ -19,12 +19,20 @@ def get_queryset(self):
         Return task logs in workspace
         """
         task_status = self.request.query_params.getlist('status')
+        expense_group_ids = self.request.query_params.get('ids')
```
ashwin1111, line 22: where are you using this ```expense_group_ids```

NileshPant1999: I will add it as a filter in if block

apps/tasks/views.py
```diff
@@ -19,12 +19,28 @@ def get_queryset(self):
         Return task logs in workspace
         """
         task_status = self.request.query_params.getlist('status')
+        ids = self.request.query_params.get('ids').split(',')
```
ashwin1111, line 22: expense_group_ids = self.request.query_params.getlist('expense_group_ids')

ashwin1111: line 25-30 not needed if you do this

apps/tasks/views.py
```diff
@@ -19,12 +19,28 @@ def get_queryset(self):
...
+        try:
+            for i in range(0, len(ids)):
+                expense_group_ids.append(int(ids[i])) 
+        except:
+            expense_group_ids = []
+            
+        
 
         if len(task_status) == 1 and task_status[0] == 'ALL':
+            task_status = ["IN_PROGRESS"]
```
ashwin1111, line 35: this is wrong no, why are you constructing only ```'IN_PROGRESS'``` even though you received ```ALL``` in query params

apps/tasks/views.py
```diff
@@ -19,12 +19,28 @@ def get_queryset(self):
...
+            task_type = 'FETCHING_EXPENSES'
+        else:
+            task_type = 'FAILED'
             task_status = ['IN_PROGRESS', 'FAILED', 'COMPLETE']
 
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
-                                           status__in=task_status).order_by('-updated_at').all()
+        task_logs = TaskLog.objects.filter(~Q(type=task_type),
+                                           workspace_id=self.kwargs['workspace_id'],
+                                           status__in=task_status,expense_group__in=expense_group_ids).order_by('-updated_at').all()
```
ashwin1111, line 43: what happens when you don't receive expense_group_ids? This will always return [] right

NileshPant1999: yes it is returning empty list

apps/tasks/views.py
```diff
@@ -19,12 +19,23 @@ def get_queryset(self):
         Return task logs in workspace
         """
         task_status = self.request.query_params.getlist('status')
-
+        ids = self.request.query_params.get('ids')
```
ashwin1111, line 22: ```getlist```

ashwin1111: 24-27

apps/tasks/views.py
```diff
@@ -19,13 +19,20 @@ def get_queryset(self):
...
-            task_status = ['IN_PROGRESS', 'FAILED', 'COMPLETE']
-
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
+        
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
+            expense_group_ids = expense_group_ids.split(',')
+            return TaskLog.objects.filter(
+                ~Q(type__in=['FETCHING_EXPENSES' ,'CREATING_VENDOR_PAYMENT']),
```
Sravanksk, line 28: can remove this `CREATING_VENDOR_PAYMENT`

apps/tasks/views.py
```diff
@@ -19,13 +19,19 @@ def get_queryset(self):
...
         """
         task_status = self.request.query_params.getlist('status')
-
-        if len(task_status) == 1 and task_status[0] == 'ALL':
-            task_status = ['IN_PROGRESS', 'FAILED', 'COMPLETE']
-
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
```
Shwetabhk, line 24: The API should be generic, this looks really hacky, It should receive query parameters generically and not according to the client, this needs to be fixed in a better way

apps/tasks/views.py
```diff
@@ -19,13 +19,19 @@ def get_queryset(self):
...
+            expense_group_ids = expense_group_ids.split(',')
```
Shwetabhk, line 25: Why are we splitting expense group ids by `,` why are we not sending them as a list from the client?

apps/tasks/views.py
```diff
@@ -19,13 +19,19 @@ def get_queryset(self):
...
+            return TaskLog.objects.filter(
+                ~Q(type__in=['FETCHING_EXPENSES']),
```
Shwetabhk, line 27: API should not be Filtering by `FETCHING_EXPENSES` client should send a request in a way that API knows what to filter

apps/tasks/views.py
```diff
@@ -19,13 +19,19 @@ def get_queryset(self):
         Return task logs in workspace
         """
         task_status = self.request.query_params.getlist('status')
-
-        if len(task_status) == 1 and task_status[0] == 'ALL':
-            task_status = ['IN_PROGRESS', 'FAILED', 'COMPLETE']
-
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
```
Shwetabhk, line 22: query_params.getlist('expense_group_ids')

apps/tasks/views.py
```diff
@@ -19,13 +19,19 @@ def get_queryset(self):
         Return task logs in workspace
         """
         task_status = self.request.query_params.getlist('status')
-
-        if len(task_status) == 1 and task_status[0] == 'ALL':
```
Shwetabhk, line 23: The same needs to be done for type

apps/tasks/views.py
```diff
@@ -19,13 +19,19 @@ def get_queryset(self):
...
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
+            expense_group_ids = expense_group_ids.split(',')
+            return TaskLog.objects.filter(
+                ~Q(type__in=['FETCHING_EXPENSES']),
+                workspace_id=self.kwargs['workspace_id'],
+                status='IN_PROGRESS',
+                expense_group__in=expense_group_ids
```
Shwetabhk, line 30: Check out kwargs in django ORM

apps/tasks/views.py
```diff
@@ -19,13 +19,23 @@ def get_queryset(self):
...
-                                           status__in=task_status).order_by('-updated_at').all()
-        return task_logs
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
+            expense_group_ids = expense_group_ids.split(',')
+            filters = {
+                'workspace_id':self.kwargs['workspace_id'],
+                'status__in': ['IN_PROGRESS'],
+                'type__in':['CREATE_BILL','CREATE_CHECK'],
```
Sravanksk, line 29: This should be 
`'type__in': ['CREATING_BILL', 'CREATING_CHECK'],`

NileshPant1999: okay will add them also

apps/tasks/views.py
```diff
@@ -19,13 +19,23 @@ def get_queryset(self):
...
-                                           status__in=task_status).order_by('-updated_at').all()
-        return task_logs
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
+            expense_group_ids = expense_group_ids.split(',')
+            filters = {
+                'workspace_id':self.kwargs['workspace_id'],
+                'status__in': ['IN_PROGRESS'],
+                'type__in':['CREATE_BILL','CREATE_CHECK','CREATING_CREDIT_CARD_PURCHASE','CREATING_JOURNAL_ENTRY'],
```
Shwetabhk, line 29: If we're sending expense group ids do we even need `type`? @NileshPant1999 @Sravanksk

apps/tasks/views.py
```diff
@@ -19,13 +19,23 @@ def get_queryset(self):
...
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
-                                           status__in=task_status).order_by('-updated_at').all()
-        return task_logs
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
+            expense_group_ids = expense_group_ids.split(',')
+            filters = {
+                'workspace_id':self.kwargs['workspace_id'],
+                'status__in': ['IN_PROGRESS'],
```
Shwetabhk, line 28: This is again not correct, we're hard coding task status
We should be sending this as a query param from the front end @NileshPant1999 @Sravanksk

apps/tasks/views.py
```diff
@@ -19,13 +19,23 @@ def get_queryset(self):
...
-
-        if len(task_status) == 1 and task_status[0] == 'ALL':
-            task_status = ['IN_PROGRESS', 'FAILED', 'COMPLETE']
-
-        task_logs = TaskLog.objects.filter(workspace_id=self.kwargs['workspace_id'],
-                                           status__in=task_status).order_by('-updated_at').all()
-        return task_logs
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+
+        if len(task_status) == 1 and task_status[0] == 'ALL' and expense_group_ids:
```
Shwetabhk, line 24: This is not correct, here we're checking if task_status  = 'ALL' and then in the if condition filtering on `IN_PROGRESS`

@NileshPant1999 @Sravanksk

apps/tasks/views.py
Shwetabhk, line 28: we should not be hard coding filters by any means, they should always come as query params

apps/tasks/views.py
```diff
@@ -1,7 +1,7 @@
 from rest_framework import generics
 from rest_framework.response import Response
 from rest_framework.views import status
-
+from django.db.models import Q
```
ashwin1111, line 4: remove this if this is not used

apps/tasks/views.py
```diff
@@ -19,13 +19,25 @@ def get_queryset(self):
...
-                                           status__in=task_status).order_by('-updated_at').all()
-        return task_logs
+        expense_group_ids = self.request.query_params.get('expense_group_ids')
+        task_type = self.request.query_params.get('task_type')
+
+        if len(task_status) == 1 and task_status[0] == 'IN_PROGRESS' and expense_group_ids:
+            expense_group_ids = expense_group_ids.split(',')
+            task_type = task_type.split(',')
+            filters = {
+                'workspace_id':self.kwargs['workspace_id'],
```
ashwin1111, line 29: one space after `:`
same for all keys below

ashwin1111 (approved):

Shwetabhk (approved):

Sravanksk (approved):

PR #35, opened 2021-01-25
Clean up tasks on exports polling

KirtiGautam:
- Updated expense group creation method to either update previously created objects in import or create new only if it doesn't exist.
- Wrote a script to delete redundant entries in the database.

### DIFF
--- a/apps/fyle/tasks.py
+++ b/apps/fyle/tasks.py
@@ -31,11 +31,11 @@ def schedule_expense_group_creation(workspace_id: int, user: str):
     jobs = fyle_sdk_connection.Jobs
     user_profile = fyle_sdk_connection.Employees.get_my_profile()['data']
 
-    task_log = TaskLog.objects.create(
+    task_log, created = TaskLog.objects.get_or_create(
         workspace_id=workspace_id,
         type='FETCHING_EXPENSES',
-        status='IN_PROGRESS'
     )
+    task_log.status = 'IN_PROGRESS'
 
     created_job = jobs.trigger_now(
         callback_url='{0}{1}'.format(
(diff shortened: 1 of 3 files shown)


apps/fyle/tasks.py
```diff
@@ -31,11 +31,11 @@ def schedule_expense_group_creation(workspace_id: int, user: str):
     jobs = fyle_sdk_connection.Jobs
     user_profile = fyle_sdk_connection.Employees.get_my_profile()['data']
 
-    task_log = TaskLog.objects.create(
+    task_log, created = TaskLog.objects.get_or_create(
```
ashwin1111, line 34: We need update or create here.

ashwin1111 (changes requested):

apps/fyle/tasks.py
ashwin1111: ignore, didn't notice the  .save() part. this is fine.

ashwin1111 (approved):

apps/fyle/tasks.py
KirtiGautam: There's a save method called below. It will update the object as soon as it hits save. We don't need to run the update method twice.

KirtiGautam: Okay 😊

scripts/004-task-cleanup.sql
```diff
@@ -0,0 +1,5 @@
+-- Script to remove redundant entries in task table
+rollback;
+begin;
+
+Delete from task_logs where type='FETCHING_EXPENSES' and status='COMPLETE';
```
Shwetabhk, line 5: Change `Delete` to `delete`

Shwetabhk: Delete would also work, but we use only lowercase for SQL keywords

Shwetabhk (approved):

scripts/004-task-cleanup.sql
KirtiGautam: Okay, making the required changes.

PR #37, opened 2021-01-27
BugFix: active_only flag, get all projects, remove special characters in accounts and customers

ashwin1111:


apps/quickbooks_online/utils.py
```diff
@@ -161,7 +161,7 @@ def sync_customers(self):
             customer_attributes.append({
                 'attribute_type': 'CUSTOMER',
                 'display_name': 'customer',
-                'value': customer['FullyQualifiedName'],
+                'value': customer['FullyQualifiedName'].replace('/', '-'),
```
ashwin1111, line 164: have to check this

Shwetabhk (approved):

PR #38, opened 2021-02-03
Migrate from Fyle Jobs to Django Q

ashwin1111:


apps/workspaces/migrations/0007_auto_20210203_1146.py
```diff
@@ -0,0 +1,32 @@
...
+        ),
+        migrations.AddField(
+            model_name='workspaceschedule',
+            name='schedule',
+            field=models.OneToOneField(null=True, on_delete=django.db.models.deletion.PROTECT, to='django_q.Schedule'),
+        ),
+        migrations.AddField(
+            model_name='workspaceschedule',
+            name='workspace',
+            field=models.OneToOneField(default=1, help_text='Reference to Workspace model', on_delete=django.db.models.deletion.PROTECT, to='workspaces.Workspace'),
```
ashwin1111, line 27: default=1 is done intentionally, removed default property in next migration file

ashwin1111: since this table already exists, we cannot add workspace column with not null. Can write a script to delete all rows in workspace_schedules and re create from UI

ashwin1111: Q: is the delete script needed to be checked into the PR?

Shwetabhk: Can just check it in for reference
` write a script to delete all rows in workspace_schedules and re create from UI`
Approach lgtm

Shwetabhk (approved):

Sravanksk (approved):

PR #40, opened 2021-02-06
Log exceptions on expense attribute does not exist

ashwin1111:


ashwin1111:
taken care here https://github.com/fylein/fyle-qbo-api/commit/2bc430559bff2fbd3890a5a3b42d87de271f568b

PR #48, opened 2021-02-15
Fix For Multiple Attachments upload failing in QBO Integration

NileshPant1999:


apps/fyle/utils.py
```diff
@@ -273,13 +273,17 @@ def get_attachments(self, expense_ids: List[str]):
         Get attachments against expense_ids
         """
         attachments = []
+        attachment_name = []
```
ashwin1111, line 276: Please check for this case.
Assume img-1.png, img-2.png for expense 1 and img-1.png, img-2.png for expense 2.
This would probably fail.

NileshPant1999: oh yes i will put this inside for loop for expense_ids

Sravanksk (approved):

apps/fyle/utils.py
```diff
@@ -275,11 +275,15 @@ def get_attachments(self, expense_ids: List[str]):
         attachments = []
         if expense_ids:
             for expense_id in expense_ids:
+                attachment_name = []
```
Shwetabhk, line 278: Rename to `attached_file_names`

Shwetabhk (approved):

PR #50, opened 2021-02-22
Category sync versioning, Auto Map support in Projects and Categories, Exclude html attachments to be posted to QBO

ashwin1111:


Shwetabhk (approved):

Shwetabhk (approved):

apps/quickbooks_online/utils.py
```diff
@@ -61,17 +61,19 @@ def sync_accounts(self, account_type: str):
             attribute_type = 'ACCOUNTS_PAYABLE'
             display_name = 'Accounts Payable'
 
-        for account in accounts:
+        category_sync_version = WorkspaceGeneralSettings.objects.get(workspace_id=self.workspace_id).category_sync_version
```
Shwetabhk, line 64: this will fail if General settings are not completed

Shwetabhk: do `.first()`

Shwetabhk:
```python
category_sync_version = 'v2'
general_settings = WorkspaceGeneralSettings.objects.filter(workspace_id=self.workspace_id).first()
if general_settings:
    category_sync_version = general_settings.category_sync_version
```

Shwetabhk: ^ This will change it to  v1 when required

apps/quickbooks_online/utils.py
```diff
@@ -61,17 +61,19 @@ def sync_accounts(self, account_type: str):
...
                 'display_name': display_name,
-                'value': unidecode.unidecode(u'{0}'.format(account['Name'])),
+                'value': unidecode.unidecode(u'{0}'.format(
+                    account['Name'] if category_sync_version == 'v1' else account['FullyQualifiedName'])),
                 'destination_id': account['Id'],
                 'active': account['Active'],
                 'detail': {
                     'fully_qualified_name': account['FullyQualifiedName']
-                }
+                } if category_sync_version == 'v1' else None
```
Shwetabhk, line 76: no need to add condition, can leave it as is

Shwetabhk (approved):

PR #51, opened 2021-02-23
Auto create Employee / Vendor

Shwetabhk:


apps/quickbooks_online/tasks.py
```diff
@@ -50,6 +50,68 @@ def load_attachments(qbo_connection: QBOConnector, ref_id: str, ref_type: str, e
...
+            if employee_mapping_setting == 'EMPLOYEE':
+                created_entity: DestinationAttribute = qbo_connection.post_employee(source_employee)
+            else:
+                created_entity: DestinationAttribute = qbo_connection.post_vendor(source_employee)
+
+            Mapping.create_or_update_mapping(
+                source_type='EMPLOYEE',
+                source_value=expense_group.description.get('employee_email'),
+                destination_type=employee_mapping_setting,
+                destination_id=created_entity.destination_id,
```
ashwin1111, line 84: I think we need to sync employee / vendor so that the created vendor/employee would reflect in destination table 🤔

apps/quickbooks_online/tasks.py
```diff
@@ -50,6 +50,68 @@ def load_attachments(qbo_connection: QBOConnector, ref_id: str, ref_type: str, e
...
+            logger.error(exception.response)
+
+            if employee_mapping_setting == 'EMPLOYEE':
+                qbo_connection.sync_employees()
+            else:
+                qbo_connection.sync_vendors()
+
+            error_response = json.loads(exception.response)['Fault']['Error'][0]
+
+            if error_response['code'] == '6240':
```
ashwin1111, line 98: what does this error mean

apps/quickbooks_online/tasks.py
```diff
@@ -50,6 +50,68 @@ def load_attachments(qbo_connection: QBOConnector, ref_id: str, ref_type: str, e
...
+                qbo_entity = DestinationAttribute.objects.get(
+                    value=source_employee.detail['full_name'],
```
ashwin1111, line 100: this would fail when any org has 2 employees of same name in qbo no?

apps/quickbooks_online/tasks.py
ashwin1111: can ignore this, noticed `post_vendor()`

apps/quickbooks_online/tasks.py
Shwetabhk: This error code comes up when employee with Display Name already exists

apps/quickbooks_online/tasks.py
```diff
@@ -151,7 +225,7 @@ def create_bill(expense_group, task_log):
         logger.error('Something unexpected happened workspace_id: %s %s', task_log.workspace_id, task_log.detail)
 
 
-def __validate_expense_group(expense_group: ExpenseGroup):
+def __validate_expense_group(expense_group: ExpenseGroup, general_settings: WorkspaceGeneralSettings):
```
ashwin1111, line 228: We should be skipping employee mapping check if they enable this feature right?
Are they handled?

apps/quickbooks_online/tasks.py
```diff
@@ -85,18 +153,24 @@ def schedule_bills_creation(workspace_id: int, expense_group_ids: List[str]):
...
 
 def create_bill(expense_group, task_log):
+    general_settings = WorkspaceGeneralSettings.objects.get(workspace_id=expense_group.workspace_id)
+
     try:
+        qbo_credentials = QBOCredential.objects.get(workspace_id=expense_group.workspace_id)
+
+        qbo_connection = QBOConnector(qbo_credentials, expense_group.workspace_id)
+
+        if expense_group.fund_source == 'PERSONAL' and general_settings.auto_map_employees \
```
ashwin1111, line 163: why are we having `expense_group.fund_source == 'PERSONAL'` this check here and not on other 3 dimensions like `JOURNAL ENTRY, CHECK` etc

apps/quickbooks_online/tasks.py
Shwetabhk: We're creating mapping in auto create too no, no reason to skip it

apps/quickbooks_online/tasks.py
Shwetabhk: Because for ccc bills we don't need to create vendor, there will be a default vendor

ashwin1111 (approved):

apps/quickbooks_online/utils.py
```diff
@@ -170,6 +205,40 @@ def sync_employees(self):
...
+        :param employee: employee attribute to be created
+        :return: Employee Desination Atribute
+        """
+        quickbooks_display_name = employee.detail['employee_code'] if (
+                auto_map_employee_preference == 'EMPLOYEE_CODE' and employee.detail['employee_code']
+        ) else employee.detail['full_name']
+
+        employee = {
+            'GivenName': employee.detail['full_name'].split(' ')[0],
+            'FamilyName': employee.detail['full_name'].split(' ')[0]
```
Shwetabhk, line 221: need to fix Index here

Sravanksk (approved):

PR #63, opened 2021-03-17
Avoid duplicate job run when the task queue dies or stuck

ashwin1111:


apps/quickbooks_online/tasks.py
```diff
@@ -3,6 +3,7 @@
 import traceback
```
ashwin1111, line 3: req from chrome console
```
for (let i=0;i<100;i++) {
fetch("http://localhost:8000/api/workspaces/1/qbo/bills/trigger/", {
  "headers": {
    "accept": "application/json, text/plain, */*",
    "accept-language": "en-GB,en-US;q=0.9,en;q=0.8",
    "authorization": "Bearer <token>",
    "content-type": "application/json",
    "sec-ch-ua": "\"Google Chrome\";v=\"89\", \"Chromium\";v=\"89\", \";Not A Brand\";v=\"99\"",
    "sec-ch-ua-mobile": "?0",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-fetch-site": "same-site"
  },
  "referrer": "http://localhost:4200/",
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": "{\"expense_group_ids\":[13]}",
  "method": "POST",
  "mode": "cors"
});
}

File: apps/quickbooks_online/tasks.py [hunk 2]
```diff
@@ -151,29 +152,46 @@ def schedule_bills_creation(workspace_id: int, expense_group_ids: List[str]):
     """
     if expense_group_ids:
         expense_groups = ExpenseGroup.objects.filter(
+            Q(tasklog__id__isnull=True) | ~Q(tasklog__status='IN_PROGRESS'),
```
Shwetabhk, line 155: Why are we not filtering out `ENQUED` here?

Shwetabhk: if task log exists and status is ENQUEUED, it will be returned from this query

ashwin1111: will add that

Shwetabhk (approved):

PR #82, opened 2021-04-09
Bulk Mapping Function example

Shwetabhk:


apps/mappings/tasks.py
```diff
@@ -17,6 +17,43 @@
...
+    :param project_attributes: Destination Attributes List
+    :param workspace_id: workspace_id
+    :return: mappings list
+    """
+    project_names = []
+
+    for project in project_attributes:
+        project_names.append(project.value)
+
+    source_attributes: List[ExpenseAttribute] = ExpenseAttribute.objects.filter(
```
Shwetabhk, line 32: This query will get all the source attributes which are unmapped and in the list of project_names

Shwetabhk: Query 1

apps/mappings/tasks.py
```diff
@@ -17,6 +17,43 @@
...
+        value__in=project_names, workspace_id=workspace_id, mapping__source_id_isnull=True).all()
+
+    source_value_id_map = {}
+
+    for source_attribute in source_attributes:
```
Shwetabhk, line 37: creating a map of all the value and id

apps/mappings/tasks.py
```diff
@@ -17,6 +17,48 @@
...
+    Bulk create mappings
+    :param destination_type: Destination Type
+    :param source_type: Source Type
+    :param destination_attributes: Destination Attributes List
+    :param workspace_id: workspace_id
+    :return: mappings list
+    """
+    attribute_value_list = []
+
+    for destination_attribute in destination_attributes:
```
ashwin1111, line 32: `attribute_value_list = destination_attributes.values_list('value', flat=True)`

apps/mappings/tasks.py
```diff
@@ -17,6 +17,48 @@
...
+        value__in=attribute_value_list, workspace_id=workspace_id, mapping__source_id__isnull=True).all()
+
+    source_value_id_map = {}
+
+    for source_attribute in source_attributes:
+        source_value_id_map[source_attribute.value.lower()] = source_attribute.id
+
+    mapping_batch = []
+
+    for detination_attribute in destination_attributes:
```
ashwin1111, line 45: `destination_attribute`

apps/mappings/tasks.py
Shwetabhk: This we can do when destination_attributes will be a queryset, right now it is not

Shwetabhk: right now it is just a list

ashwin1111 (approved):

apps/mappings/tasks.py
```diff
@@ -17,6 +17,47 @@
 logger = logging.getLogger(__name__)
 
 
+def bulk_create_mappings(destination_attributes: List[DestinationAttribute], source_type: str,
```
ashwin1111, line 20: `destination_attributes: list`

apps/mappings/tasks.py
```diff
@@ -235,36 +242,9 @@ def auto_create_category_mappings(workspace_id):
...
-                    'source_value': category.value,
-                    'destination_value': category.value
-                }
-                logger.error(
-                    'Error while creating categories auto mapping workspace_id - %s %s',
-                    workspace_id, {'payload': detail}
-                )
-                raise ExpenseAttribute.DoesNotExist
-
+        category_mappings = bulk_create_mappings(fyle_categories, 'CATEOGORY', 'ACCOUNT', workspace_id)
```
Sravanksk, line 247: typo - `CATEGORY`

Sravanksk (approved):

ashwin1111 (approved):

PR #85, opened 2021-04-28
move dashboard sync calls to backend

NileshPant1999:


apps/quickbooks_online/utils.py
```diff
@@ -161,6 +161,7 @@ def sync_vendors(self):
 
         DestinationAttribute.bulk_create_or_update_destination_attributes(
             vendor_attributes, 'VENDOR', self.workspace_id, True)
+
```
Sravanksk, line 164: remove extra line

apps/quickbooks_online/views.py
```diff
@@ -710,3 +711,69 @@ def post(self, request, *args, **kwargs):
             data={},
             status=status.HTTP_200_OK
         )
+
+
+class SyncQuickbooksDimensionView(generics.ListCreateAPIView):
+    """
+    Sync Quickbooks Intacct Dimension View
```
Sravanksk, line 718: remove `Intacct`

Sravanksk (approved):

apps/quickbooks_online/utils.py
```diff
@@ -270,6 +273,60 @@ def sync_customers(self):
             customer_attributes, 'CUSTOMER', self.workspace_id, True)
         return []
 
+
+    def sync_dimensions(self):
+
```
ashwin1111, line 278: rm

apps/quickbooks_online/utils.py
```diff
@@ -270,6 +273,60 @@ def sync_customers(self):
...
+        try:
+            self.sync_accounts('Accounts Payable')
```
ashwin1111, line 280: Please refactor sync_accounts. We are appending same values with different display and attribute. We can achieve the same by a single call, instead of calling qbo api 5 times.

apps/quickbooks_online/utils.py
```diff
@@ -59,57 +61,82 @@ def get_or_create_vendor(self, vendor_name: str, email: str = None, create: bool
...
-            if account_type == 'Expense':
-                attribute['value'] = attribute['value'].replace('/', '-')
-
-            account_attributes.append(attribute)
+            if account['AccountType'] == 'Expense':
+                account_attributes['account'].append({
+                        'attribute_type': 'ACCOUNT',
+                        'display_name': 'Account',
+                        'value': unidecode.unidecode(u'{0}'.format(
+                            account['Name'] if category_sync_version == 'v1' else account['FullyQualifiedName'])).replace('/', '-'),
```
ashwin1111, line 88: ```account['Name'].replace('/', '-') if category_sync_version == 'v1' else account['FullyQualifiedName'])).replace('/', '-')```

Please verify this and change. Even for account['Name'] we should be replacing '/'

PR #115, opened 2021-07-22
Move employee mapping setting to general settings

ashwin1111:


apps/workspaces/utils.py
```diff
@@ -91,6 +94,7 @@ def create_or_update_general_settings(general_settings_payload: Dict, workspace_
         workspace_id=workspace_id,
         category_sync_version=category_sync_version,
         defaults={
+            'employee_field_mapping': general_settings_payload['employee_field_mapping'],
```
ashwin1111, line 97: TODO: add script

Shwetabhk:
@ashwin1111 Please remove this file `cache.db-journal`
Also add it to git ignore

Shwetabhk (approved):

PR #121, opened 2021-07-28
Fixes in employee mappings migration

Shwetabhk:


ashwin1111:
Taken care here https://github.com/fylein/fyle-qbo-api/pull/122/

PR #122, opened 2021-07-28
Employee Mappings fixes post testing

ashwin1111:


sql/scripts/012-generate-employee-mapping.sql
```diff
@@ -2,9 +2,10 @@
 rollback;
 begin;
 
-insert into employee_mappings
```
ashwin1111, line 5: TODO: update delete workspace function

apps/mappings/tasks.py
```diff
@@ -357,25 +357,33 @@ def create_mappings_and_update_flag(mapping_creation_batch: List[EmployeeMapping
     if mapping_creation_batch:
         created_mappings = EmployeeMapping.objects.bulk_create(mapping_creation_batch, batch_size=50)
         mappings.extend(created_mappings)
+        print('created_mappings',created_mappings)
```
ashwin1111, line 360: rm loggers before merging

PR #128, opened 2021-08-16
adding support to export CCC credits as credit card credits to QBO

Sravanksk:


apps/quickbooks_online/models.py
```diff
@@ -666,6 +666,157 @@ def create_credit_card_purchase_lineitems(expense_group: ExpenseGroup):
...
+    """
+    QBO CreditCardCredit Lineitem
+    """
+    id = models.AutoField(primary_key=True)
+    credit_card_credit = models.ForeignKey(CreditCardCredit, on_delete=models.PROTECT,
+                                           help_text='Reference to credit card credit')
+    expense = models.OneToOneField(Expense, on_delete=models.PROTECT, help_text='Reference to Expense')
+    account_id = models.CharField(max_length=255, help_text='QBO account id')
+    class_id = models.CharField(max_length=255, help_text='QBO class id', null=True)
+    customer_id = models.CharField(max_length=255, help_text='QBO customer id', null=True)
```
Shwetabhk, line 765: `refund = True` add a flag in purchase table

apps/quickbooks_online/models.py
```diff
@@ -666,6 +666,157 @@ def create_credit_card_purchase_lineitems(expense_group: ExpenseGroup):
...
+        general_mappings = GeneralMapping.objects.get(workspace_id=expense_group.workspace_id)
+        workspace_general_settings = WorkspaceGeneralSettings.objects.get(workspace_id=expense_group.workspace_id)
+        employee_field_mapping = workspace_general_settings.employee_field_mapping
+        entity_id = None
+
+        department_id = get_department_id_or_none(expense_group)
+
+        private_note = construct_private_note(expense_group)
+
+        if map_merchant_to_vendor:
```
ashwin1111, line 707: line 707 - 721 is exactly same as CCP, can we try to put them into a generic function?

ashwin1111: okay, I see whole of this method is duplicated, is it possible to have a generic one for generating object to upsert?

PR #129, opened 2021-08-19
adding support to export CCC credits as credit card credits to QBO

Sravanksk:


apps/quickbooks_online/utils.py
```diff
@@ -579,9 +580,15 @@ def __construct_credit_card_purchase(self, credit_card_purchase: CreditCardPurch
         :return: constructed credit_card_purchase
         """
         line = self.__construct_credit_card_purchase_lineitems(credit_card_purchase_lineitems)
```
Sravanksk, line 582: Assuming we get a single item in the list since `CREDIT_CARD_PURCHASES` are grouped by `expense_id`

Shwetabhk (approved):

PR #130, opened 2021-08-20
sentry integration changes

labhvam5:


fyle_qbo_api/settings.py
```diff
@@ -258,3 +258,9 @@
 CACHE_EXPIRY = 3600
 
 CORS_ORIGIN_ALLOW_ALL = True
+
+CORS_ALLOW_HEADERS = [
```
abhishek1234321, line 262: have we set this somewhere in code `Access-Control-Allow-Origin: *`?

abhishek1234321: could we check this https://github.com/getsentry/sentry-javascript/issues/2859#issuecomment-682644114

also, did you try using the official account on staging? Let me know if you need help in changing Settings in our official account

But I would recommend to try out a few settings in your account

abhishek1234321 (changes requested):

PR #133, opened 2021-09-02
post expenses even when the accounting period is closed

NileshPant1999:


apps/quickbooks_online/utils.py
```diff
@@ -427,9 +429,24 @@ def post_bill(self, bill: Bill, bill_lineitems: List[BillLineitem]):
...
-        created_bill = self.connection.bills.post(bills_payload)
-        return created_bill
+        try:
+            bills_payload = self.__construct_bill(bill, bill_lineitems)
+            created_bill = self.connection.bills.post(bills_payload)
+            return created_bill
+
+        except WrongParamsError as bad_request:
+            general_settings = WorkspaceGeneralSettings.objects.filter(workspace_id=self.workspace_id).first()
+            error_response = json.loads(bad_request.response)['Fault']['Error'][0]
```
ashwin1111, line 439: @NileshPant1999 bring this line inside if. Also, few questions
1. Do we get `['Fault']['Error']` in the error response for all kind of errors?
2. If yes, will it be a list every time?
3. if yes, will it have the 1st element every time for any kind of errors?

ashwin1111: Also can we do it this way,
```
if general_settings.forced_export:
    error_response = json.loads(bad_request.response)['Fault']['Error'][0]
    if error_response['Message'] == 'Account Period Closed' and error_response['code'] == '6200':
        # post_to_qbo()
else:
    raise

File: apps/quickbooks_online/utils.py [hunk 2]
```diff
@@ -427,9 +429,24 @@ def post_bill(self, bill: Bill, bill_lineitems: List[BillLineitem]):
...
+            
+            if general_settings.forced_export and error_response['Message'] == 'Account Period Closed' and error_response['code'] == '6200':
+                book_closed_date = self.connection.preferences.get()['AccountingInfoPrefs']['BookCloseDate']
+                txn_date = datetime.strptime(book_closed_date, '%Y-%m-%d') + timedelta(days=1)
+                bills_payload['TxnDate'] = txn_date.strftime("%Y-%m-%d")
```
ashwin1111, line 444: Do you think we need to store the updated date in db? either in exports* tables (not user visible) or expense groups(user visible)?
@Shwetabhk thoughts?

ashwin1111 (commented):
Added few questions and suggestions, those will apply for all kind of exports.

Sravanksk (approved):

apps/quickbooks_online/utils.py
NileshPant1999: yes ['Fault']['Error'] is present in every response.
yes it is list every time and it is the first element every time.

ashwin1111 (approved):

apps/quickbooks_online/utils.py
```diff
@@ -423,13 +426,33 @@ def __construct_bill(self, bill: Bill, bill_lineitems: List[BillLineitem]) -> Di
...
+            if general_settings.change_accounting_period:
+                error_response = json.loads(bad_request.response)['Fault']['Error'][0]
+                if error_response['Message'] == 'Account Period Closed' and error_response['code'] == '6200':
+                    book_closed_date = self.connection.preferences.get()['AccountingInfoPrefs']['BookCloseDate']
+                    txn_date = datetime.strptime(book_closed_date, '%Y-%m-%d') + timedelta(days=1)
+
+                    bills_payload['TxnDate'] = txn_date.strftime("%Y-%m-%d")
+                    created_bill = self.connection.bills.post(bills_payload)
+
+                    bill = Bill.objects.get(expense_group=expense_group)
```
ashwin1111, line 450: we already have the bill instance, this is not needed

ashwin1111: same everywhere

ashwin1111 (changes requested):

ashwin1111 (approved):

PR #138, opened 2021-09-21
sentry integration

labhvam5:


fyle_qbo_api/sentry.py
```diff
@@ -0,0 +1,28 @@
...
+        )
+
+    @staticmethod
+    def traces_sampler(sampling_context):
+        # avoiding ready APIs in performance tracing
+        if sampling_context.get('wsgi_environ') is not None:
+            if sampling_context['wsgi_environ']['PATH_INFO'] in ['/ready']:
+                return 0
+
+        return 1.0
```
ashwin1111, line 28: Can we reduce it to 0.5? If we are not getting much we can increase this.

ashwin1111: I assume this is for traces sample rate, correct me if I am wrong

ashwin1111 (approved):

ashwin1111:
@labhvam5 make sure you do squash merge

fyle_qbo_api/sentry.py
labhvam5: yes this is the traces sample rate.

fyle_qbo_api/sentry.py
```diff
@@ -0,0 +1,29 @@
...
+    def init():
+        sentry_sdk.init(
+            dsn=os.environ.get('SENTRY_DSN'),
+            send_default_pii=True,
+            integrations=[DjangoIntegration()],
+            environment='quickbooks-api',
+            traces_sampler=Sentry.traces_sampler,
+            release=os.environ.get('RELEASE'),
+            attach_stacktrace=True,
+            request_bodies= 'small',
```
ashwin1111, line 19: `request_bodies='small'`
No spacing

PR #151, opened 2021-11-02
Default Mapping Missing error should not come when mapping is present

NileshPant1999:


apps/quickbooks_online/tasks.py
```diff
@@ -450,12 +441,22 @@ def __validate_expense_group(expense_group: ExpenseGroup, general_settings: Work
...
+            if not lineitem.tax_group_id:
+                bulk_errors.append({
+                    'row': None,
+                    'expense_group_id': expense_group.id,
+                    'value': 'Default Tax Code',
+                    'type': 'General Mapping',
+                    'message': 'Default Tax Code not found'
+                })
+
         if general_settings.import_tax_codes and lineitem.tax_group_id:
```
ashwin1111, line 454: why not add this as else part? in the above block

apps/quickbooks_online/tasks.py
```diff
@@ -450,12 +441,22 @@ def __validate_expense_group(expense_group: ExpenseGroup, general_settings: Work
...
-            tax_group  = ExpenseAttribute.objects.get(
+            tax_group = ExpenseAttribute.objects.filter(
                 workspace_id=expense_group.workspace_id,
                 attribute_type='TAX_GROUP',
                 source_id=lineitem.tax_group_id
-            )
+            ).first()
 
             tax_code = Mapping.objects.filter(
```
ashwin1111, line 461: `if tax_group:`
or may be alter this line
`source__value=tax_group.value if tax_group else lineitem. tax_group_id,`

ashwin1111 (changes requested):

PR #159, opened 2021-11-19
Feature: Added Support for Charts of Accounts

NileshPant1999:


apps/quickbooks_online/utils.py
```diff
@@ -125,20 +124,22 @@ def sync_accounts(self):
...
-                    'destination_id': account['Id'],
-                    'active': account['Active'],
-                    'detail': {
-                        'fully_qualified_name': account['FullyQualifiedName'],
-                        'account_type': account['AccountType']
-                    }
-                })
 
-            elif account['AccountType'] == 'Credit Card' and value:
+            for chart_of_account in general_settings.charts_of_accounts:
```
Shwetabhk, line 128: no need for a for loop here
can simply do `if account['AccountType'] in chart_of_accounts`

apps/workspaces/models.py
```diff
@@ -58,6 +60,10 @@ class WorkspaceGeneralSettings(models.Model):
     import_projects = models.BooleanField(default=False, help_text='Auto import projects to Fyle')
     import_tax_codes = models.BooleanField(default=False, help_text='Auto import tax codes to Fyle', null=True)
     change_accounting_period = models.BooleanField(default=False, help_text='Export Expense when accounting period is closed')
+    charts_of_accounts = ArrayField(
+        base_field=models.CharField(max_length=100), default=get_default_chart_of_accounts,
+        help_text='list of charts of accounts to be imported to fyle'
```
Shwetabhk, line 65: list of chart of account type to be imported into Fyle

apps/workspaces/models.py
```diff
@@ -43,6 +43,8 @@ class Meta:
         db_table = 'workspace_schedules'
 
 
+def get_default_chart_of_accounts():
+    return ['Expense']
```
Shwetabhk, line 47: add a line break after this

ashwin1111 (approved):

PR #166, opened 2021-12-16
use post_save signal to sync_accounts

Sravanksk:


ashwin1111 (approved):

apps/quickbooks_online/utils.py
```diff
@@ -124,8 +124,7 @@ def sync_accounts(self):
             value = format_special_characters(
                 account['Name'] if category_sync_version == 'v1' else account['FullyQualifiedName']
             )
-
-            if account['AccountType'] in general_settings.charts_of_accounts and value:
+            if general_settings and (account['AccountType'] in general_settings.charts_of_accounts and value):
```
ashwin1111, line 127: () is not required

PR #169, opened 2021-12-28
Bug Fix: Send the tax detail even if the tax amount is 0

NileshPant1999:


apps/quickbooks_online/utils.py
```diff
@@ -808,11 +808,11 @@ def __group_journal_entry_credits(self, journal_entry_lineitems: List[JournalEnt
 
         for lineitem in journal_entry_lineitems:
             total_tax = self._get_total_tax(lineitem, general_mappings)
-            if lineitem.debit_account_id in card_objects:
-                card_objects[lineitem.debit_account_id]['amount'] = card_objects[lineitem.debit_account_id]['amount'] + lineitem.amount
-                card_objects[lineitem.debit_account_id]['total_tax'] = card_objects[lineitem.debit_account_id]['total_tax'] + self._get_total_tax(lineitem, general_mappings)
+            if lineitem.debit_account_id+lineitem.tax_code in card_objects:
```
Sravanksk, line 811: fix indentation, add space before and after `+`

apps/quickbooks_online/utils.py
```diff
@@ -901,15 +901,15 @@ def __construct_journal_entry_lineitems(self, journal_entry_lineitems: List[Jour
...
+            if (line.tax_code and line.tax_amount is not None):
                 final_amount = abs(line.amount) - abs(line.tax_amount)
             else:
                 final_amount = abs(self.get_tax_inclusive_amount(line.amount, general_mappings.default_tax_code_id))
 
             lineitem = {
                 'DetailType': 'JournalEntryLineDetail',
                 'Description': line.description,
-                'Amount': final_amount,
+                'Amount': round(final_amount,2),
```
Sravanksk, line 912: indentation

Sravanksk (approved):

apps/quickbooks_online/utils.py
```diff
@@ -806,18 +806,21 @@ def __group_journal_entry_credits(self, journal_entry_lineitems: List[JournalEnt
         card_objects = {}
         data = []
 
+        configuration = WorkspaceGeneralSettings.objects.get(workspace_id=self.workspace_id)
         for lineitem in journal_entry_lineitems:
+            tax_code = lineitem.tax_code if lineitem.tax_code else general_mappings.default_tax_code_id if \
+                (configuration.import_tax_codes and not lineitem.tax_code) else ""
             total_tax = self._get_total_tax(lineitem, general_mappings)
```
Sravanksk, line 813: we are assigning `default_tax_code_id` if tax_code is null in line_items

Sravanksk: If tax_code is null, then tax_amount would also be null right. In this case, the amount will mismatch right?

NileshPant1999: no, we are not doing that in that step, we are doing that in construct payload step, will clean the code later and move all the logic to line items step.

PR #174, opened 2022-01-20
Platform migration v2 (  )

Sravanksk:


apps/fyle/utils.py
```diff
@@ -138,7 +138,7 @@ def get_fyle_orgs(self, cluser_domain):
         return self._get_request(api_url, params)
 
     def get_expenses(self, state: List[str], fund_source: List[str],
```
ashwin1111, line 140: why is this needed?

apps/fyle/views.py
```diff
@@ -446,10 +323,9 @@ def post(self, request, *args, **kwargs):
 
             if workspace.source_synced_at is None or time_interval.days > 0:
                 fyle_credentials = FyleCredential.objects.get(workspace_id=kwargs['workspace_id'])
-                fyle_connector = FyleConnector(fyle_credentials.refresh_token, kwargs['workspace_id'])
                 platform = PlatformConnector(fyle_credentials)
 
-                fyle_connector.sync_dimensions()
+                platform.import_fyle_dimensions()
                 platform.tax_groups.sync()
```
ashwin1111, line 329: this is not needed, make use the flag in previous function

apps/fyle/views.py
```diff
@@ -487,10 +363,9 @@ def post(self, request, *args, **kwargs):
         """
         try:
             fyle_credentials = FyleCredential.objects.get(workspace_id=kwargs['workspace_id'])
-            fyle_connector = FyleConnector(fyle_credentials.refresh_token, kwargs['workspace_id'])
             platform = PlatformConnector(fyle_credentials)
 
-            fyle_connector.sync_dimensions()
+            platform.import_fyle_dimensions()
             platform.tax_groups.sync()
```
ashwin1111, line 369: same

ashwin1111 (changes requested):

ashwin1111 (approved):

PR #176, opened 2022-01-24
Qbo cards mapping

Sravanksk:


apps/mappings/constants.py
```diff
@@ -41,5 +41,6 @@
     'reimbursable',
     'receipts',
     'paid date',
-    'expense created date'
+    'expense created date',
+    'corporate_card'
```
ashwin1111, line 45: didn't get it, is there any system field for this? I am able to create a select type custom field via UI for corporate card

apps/quickbooks_online/models.py
```diff
@@ -840,6 +838,35 @@ def create_journal_entry_lineitems(
...
 
+            if expense_group.fund_source == 'PERSONAL':
+                if employee_field_mapping == 'VENDOR':
+                    debit_account_id = GeneralMapping.objects.get(
+                        workspace_id=expense_group.workspace_id).accounts_payable_id
+                elif employee_field_mapping == 'EMPLOYEE':
+                    debit_account_id = GeneralMapping.objects.get(
+                        workspace_id=expense_group.workspace_id).bank_account_id
+            elif expense_group.fund_source == 'CCC':
+                if workspace_general_settings.map_fyle_cards_qbo_account:
```
ashwin1111, line 849: can you move this block of code to a separate function pliss, duplicate code

apps/workspaces/utils.py
```diff
@@ -127,10 +128,18 @@ def create_or_update_general_settings(general_settings_payload: Dict, workspace_
             'sync_fyle_to_qbo_payments': general_settings_payload['sync_fyle_to_qbo_payments'],
             'sync_qbo_to_fyle_payments': general_settings_payload['sync_qbo_to_fyle_payments'],
             'map_merchant_to_vendor': map_merchant_to_vendor,
-            'je_single_credit_line': general_settings_payload['je_single_credit_line']
+            'je_single_credit_line': general_settings_payload['je_single_credit_line'],
+            'map_fyle_cards_qbo_account': general_settings_payload['map_fyle_cards_qbo_account']
         }
     )
 
+    if not general_settings.map_fyle_cards_qbo_account:
```
ashwin1111, line 136: Didn't get this, why are we doing something like delete via code?

apps/workspaces/utils.py
```diff
@@ -127,10 +128,18 @@ def create_or_update_general_settings(general_settings_payload: Dict, workspace_
...
+        mapping_setting = MappingSetting.objects.filter(
+            workspace_id=workspace_id, source_field='CORPORATE_CARD', destination_field='CREDIT_CARD_ACCOUNT'
+        ).first()
+        if mapping_setting:
+            mapping_setting.delete()
```
ashwin1111, line 141: todo item for me - understand how are we planning to migrate existing customers

ashwin1111 (changes requested):

apps/workspaces/utils.py
Sravanksk: if configuration updates happen, we should delete mapping settings right, otherwise we end up showing Cards mapping tab

apps/workspaces/utils.py
Sravanksk: will be adding a sql script for migrating all existing customers

apps/mappings/constants.py
ashwin1111: @Sravanksk did you find out the reason behind this?

apps/quickbooks_online/models.py
```diff
@@ -68,6 +68,30 @@ def construct_private_note(expense_group: ExpenseGroup):
     return private_note
 
 
+def get_ccc_account_id(workspace_general_settings, general_mappings, expense, description):
```
ashwin1111, line 71: param and return type pleaseee

ashwin1111 (approved):

apps/mappings/constants.py
Sravanksk: fixed in this commit 
https://github.com/fylein/fyle-qbo-api/pull/176/commits/13ec536cd50781542c28d271fbbbe95c77998e1a

PR #180, opened 2022-01-31
Bug Fix: Cards mapping

Sravanksk:


apps/mappings/signals.py
```diff
@@ -29,10 +27,6 @@ def run_post_mapping_settings_triggers(sender, instance: MappingSetting, **kwarg
     if instance.is_custom:
         schedule_fyle_attributes_creation(int(instance.workspace_id))
 
-    workspace_general_settings = WorkspaceGeneralSettings.objects.filter(workspace_id=instance.workspace_id).first()
```
Sravanksk, line 32: At this point we won't have WorkspaceGeneralSettings for new orgs, so this is failing and not required

apps/workspaces/utils.py
```diff
@@ -169,7 +169,7 @@ def create_or_update_general_settings(general_settings_payload: Dict, workspace_
 
 
 def delete_cards_mapping_settings(workspace_general_settings: WorkspaceGeneralSettings):
-    if not workspace_general_settings.map_fyle_cards_qbo_account:
+    if not workspace_general_settings.map_fyle_cards_qbo_account or not workspace_general_settings.corporate_credit_card_expenses_object:
```
ashwin1111, line 172: Why do we even post this mapping setting when ccc exports are not enabled in the first place?

apps/workspaces/models.py
```diff
@@ -83,6 +83,7 @@ class WorkspaceGeneralSettings(models.Model):
     category_sync_version = models.CharField(default='v1', max_length=50, help_text='Category sync version')
     je_single_credit_line = models.BooleanField(default=False, help_text='Single Credit Line for Journal Entries')
     map_fyle_cards_qbo_account = models.BooleanField(default=True, help_text='Map Fyle Cards to QBO Accounts')
+    skip_cards_mapping = models.BooleanField(default=False, help_text='Skip cards mapping')
```
ashwin1111, line 86: Can you write a script to set this as true for onboarded orgs

apps/mappings/signals.py
```diff
@@ -29,10 +27,6 @@ def run_post_mapping_settings_triggers(sender, instance: MappingSetting, **kwarg
...
-
-    delete_cards_mapping_settings(workspace_general_settings)
```
ashwin1111, line 34: why is this removed?

apps/workspaces/utils.py
Sravanksk: This will be enabled for all orgs by default, so just to ignore one edge case

ashwin1111: Why can't you exclude them? We are inserting via sql query right?

apps/mappings/signals.py
Sravanksk: this will be called post mapping_settings, For new orgs at this point we don't have workspace_general_settings
This call is still in progress IMO, so this is throwing 500

ashwin1111: hey, why not handle them with a if condition? if POST APIs time differ then we will definitely won't be consistent since we are calling both apis in parallel from app.

apps/mappings/signals.py
ashwin1111: safely ignore them

apps/workspaces/utils.py
Sravanksk: Yes, this condition will come incase config update, just to catch that case, have added this check

ashwin1111 (approved):

PR #195, opened 2022-03-14
set qbo refresh_token column to null instead of deleting whole details

Sravanksk:


apps/workspaces/urls.py
```diff
@@ -28,7 +28,7 @@
     path('<int:workspace_id>/credentials/fyle/', ConnectFyleView.as_view({'get': 'get'}), name='get-fyle-credentials'),
     path('<int:workspace_id>/credentials/fyle/delete/', ConnectFyleView.as_view({'post': 'delete'}), name='delete-fyle-credentials'),
     path('<int:workspace_id>/connect_qbo/authorization_code/', ConnectQBOView.as_view({'post': 'post'})),
-    path('<int:workspace_id>/credentials/qbo/delete/', ConnectQBOView.as_view({'post': 'delete'}), name='delete-qbo-credentials'),
+    path('<int:workspace_id>/credentials/qbo/delete/', ConnectQBOView.as_view({'patch': 'patch'}), name='delete-qbo-credentials'),
```
ashwin1111, line 31: can we remove /delete from the path, we are doing a patch request on a delete endpoint :P

apps/workspaces/views.py
```diff
@@ -313,28 +313,38 @@ def post(self, request, **kwargs):
...
             'workspace_id': workspace_id,
-            'message': 'QBO credentials deleted'
+            'message': 'QBO Refresh Token deleted'
         })
 
     def get(self, request, **kwargs):
         """
         Get QBO Credentials in Workspace
         """
+        workspace = Workspace.objects.get(pk=kwargs['workspace_id'])
```
ashwin1111, line 332: this is not needed

apps/workspaces/views.py
```diff
@@ -313,28 +313,38 @@ def post(self, request, **kwargs):
...
+        qbo_credentials = QBOCredential.objects.get(workspace=workspace)
```
ashwin1111, line 333: `workspace_id=kwargs['workspace_id'])`

apps/workspaces/views.py
```diff
@@ -313,28 +313,38 @@ def post(self, request, **kwargs):
...
         try:
-            workspace = Workspace.objects.get(pk=kwargs['workspace_id'])
-            qbo_credentials = QBOCredential.objects.get(workspace=workspace)
+            if qbo_credentials.refresh_token:
```
ashwin1111, line 335: if else is not needed, add refresh_token not null filter directly

ashwin1111 (approved):

PR #196, opened 2022-03-16
add is_expired flag to qbo_credentials table

Sravanksk:


apps/workspaces/views.py
```diff
@@ -318,6 +320,7 @@ def patch(self, request, **kwargs):
         workspace_id = kwargs['workspace_id']
         qbo_credentials = QBOCredential.objects.get(workspace_id=workspace_id)
         qbo_credentials.refresh_token = None
+        qbo_credentials.is_expired = True
```
ashwin1111, line 323: It should be false here, in patch call, we are disconnecting the connection.

ashwin1111 (approved):

PR #202, opened 2022-03-22
mapping all QBO accounts to category dialogue in QBO Integration

DhaaraniCIT:


apps/quickbooks_online/utils.py
```diff
@@ -124,7 +124,11 @@ def sync_accounts(self):
             value = format_special_characters(
                 account['Name'] if category_sync_version == 'v1' else account['FullyQualifiedName']
             )
-            if general_settings and account['AccountType'] in general_settings.charts_of_accounts and value:
+            charts_of_accounts=['Expence', 'Other Expense', 'Fixed Asset', 'Cost of Goods Sold','Current Liability',
```
Shwetabhk, line 127: `Expense` is spelled incorrectly, please check other spellings too

Shwetabhk, line 127: Define this as a constant outside this class
```python
CHART_OF_ACCOUNTS = [...]


class QBOConnector:
    def sync_accounts():
        if general_settings and account['AccountType'] in CHART_OF_ACCOUNTS and value:
```

apps/mappings/tasks.py
```diff
@@ -242,13 +242,18 @@ def upload_categories_to_fyle(workspace_id):
...
     platform.categories.sync()
     qbo_connection.sync_accounts()
     qbo_attributes: List[DestinationAttribute] = DestinationAttribute.objects.filter(
         workspace_id=workspace_id, attribute_type='ACCOUNT').all()
-    qbo_attributes = remove_duplicates(qbo_attributes)
-
-    fyle_payload: List[Dict] = create_fyle_categories_payload(qbo_attributes, workspace_id)
+    general_settings = WorkspaceGeneralSettings.objects.filter(workspace_id=workspace_id).first()
+    qbo_attributes = list(qbo_attributes)
+    for qbo_attribute in qbo_attributes:
```
Shwetabhk, line 252: instead of using a for loop, we can directly filter in the SQL using in_clause
on line 248

```python
qbo_attributes: List[DestinationAttribute] = DestinationAttribute.objects.filter(
        workspace_id=workspace_id, attribute_type='ACCOUNT').all()
```

Modify this query so that it only returns allowed chart of accounts

Shwetabhk (changes requested):

Shwetabhk (approved):

apps/quickbooks_online/utils.py
```diff
@@ -35,6 +35,8 @@ def format_special_characters(value: str) -> str:
 
     return formatted_string
 
+charts_of_accounts=['Expense', 'Other Expense', 'Fixed Asset', 'Cost of Goods Sold','Current Liability',
+            'Equity', 'Other Current Asset', 'Other Current Liability', 'Long Term Liability', 'Current Asset']
```
Sravanksk, line 39: @DhaaraniCIT please follow python linting - this line can be modified to this - 
`charts_of_accounts = [
    'Expense', 'Other Expense', 'Fixed Asset', 'Cost of Goods Sold', 'Current Liability', 'Equity',
    'Other Current Asset', 'Other Current Liability', 'Long Term Liability', 'Current Asset'
]`

Sravanksk (approved):

PR #212, opened 2022-03-29
Adding support for Placeholder in expense field creation

Shwetabhk:


apps/workspaces/apis/import_settings/views.py
```diff
@@ -7,6 +7,9 @@
 
 class ImportSettingsView(generics.RetrieveUpdateAPIView):
     serializer_class = ImportSettingsSerializer
+    permission_classes = []
```
Shwetabhk, line 10: will remove this

Sravanksk (approved):

ashwin1111 (approved):

apps/workspaces/views.py
```diff
@@ -30,6 +31,10 @@
     WorkSpaceGeneralSettingsSerializer, WorkspaceScheduleSerializer
 from ..fyle.models import ExpenseGroupSettings
 
+logger = logging.getLogger(__name__)
```
Shwetabhk, line 34: This bug was introduced by Nilesh Sir

PR #227, opened 2022-04-08
remove mandatory field validations for expense_group_settings field in export settings API

Sravanksk:


apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -26,6 +26,9 @@ class Meta:
 
 
 class ExpenseGroupSettingsSerializer(serializers.ModelSerializer):
+    reimbursable_expense_group_fields = serializers.ListField(allow_null=True, required=False)
```
Shwetabhk, line 29: we don't need this right?

apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -26,6 +26,9 @@ class Meta:
...
+    reimbursable_export_date_type = serializers.CharField(allow_null=True, allow_blank=True, required=False)
+
```
Shwetabhk, line 31: same

Shwetabhk (approved):

PR #228, opened 2022-04-12
Cards Mapping Bug Fix: create credit lines for individual cards

Sravanksk:


ashwin1111 (approved):

apps/quickbooks_online/utils.py
```diff
@@ -797,41 +796,61 @@ def __group_journal_entry_credits(self, journal_entry_lineitems: List[JournalEnt
...
+                    'Entity': {
+                        'EntityRef': {
+                            'value': line['entity_id']
+                        },
+                        'Type': line['entity_type'],
+                    },
+                    'TaxCodeRef': {
+                        'value': line['tax_code'] if line['tax_code'] else general_mappings.default_tax_code_id
+                    },
+                    'TaxInclusiveAmt': total_sum,
```
labhvam5, line 848: The 'TaxInclusiveAmt': total_sum or should this be  'TaxInclusiveAmt':line['amount']?

apps/quickbooks_online/utils.py
```diff
@@ -792,28 +792,33 @@ def post_credit_card_purchase(self, credit_card_purchase: CreditCardPurchase,
...
-        total_sum = sum(line.amount for line in journal_entry_lineitems)
+    def _get_total_tax(self, lineitem, general_mappings):
         total_tax = 0
+        if lineitem.tax_code and lineitem.tax_amount:
+            total_tax += lineitem.tax_amount
+        else:
+            total_tax += round(
+                lineitem.amount - self.get_tax_inclusive_amount(lineitem.amount, general_mappings.default_tax_code_id), 2
+            )
+        return total_tax
```
NileshPant1999, line 803: @Sravanksk I think we need to consider a case where there can be +ve and -ve line_items in the same report, in that case, this will fail.

labhvam5: @NileshPant1999 @Sravanksk i dont think this will create a problem since the function you have written will only get +ve lineitems. If you check the __construct_journal_entry_lineitems() there we are splitting the +ve and -ve ones and only sending +ve to the __construct_journal_entry_lineitems(). Which in turn will send it to _get_total_tax(). So a mix of +ve and -ve lineitems wont be sent . @NileshPant1999 if only +ve lineitems are sent then i dont think it will break. please confirm?

NileshPant1999 (approved):

labhvam5 (approved):

PR #242, opened 2022-04-27
Update reimbursable export to null on employee field mapping update and date filter bug fix

Sravanksk:


apps/workspaces/apis/map_employees/serializers.py
```diff
@@ -30,6 +30,13 @@ def update(self, instance, validated_data):
         workspace_id = instance.id
         workspace_general_settings = validated_data.pop('workspace_general_settings')
 
+        workspace_general_settings_instance = WorkspaceGeneralSettings.objects.filter(workspace_id=instance.id).first()
+
+        if workspace_general_settings_instance.employee_field_mapping != workspace_general_settings['employee_field_mapping']:
```
ashwin1111, line 35: `if workspace_general_settings_instance` and blabla. For first save, this might break

ashwin1111 (approved):

PR #250, opened 2022-04-29
create cards_mappings based on configurations in export_settings API 

Sravanksk:


apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -135,17 +140,36 @@ def update(self, instance, validated):
...
                     'corporate_credit_card_expenses_object'),
                 'map_merchant_to_vendor': map_merchant_to_vendor,
-                'category_sync_version': category_sync_version
+                'category_sync_version': category_sync_version,
+                'map_fyle_cards_qbo_account': enable_cards_mapping
             }
         )
 
+        if enable_cards_mapping:
+            with transaction.atomic():
```
ashwin1111, line 162: transaction block is not needed

ashwin1111 (approved):

PR #263, opened 2022-05-23
Adding Tax Error to errors table for resolution on UI

Shwetabhk:


apps/quickbooks_online/tasks.py
```diff
@@ -574,6 +574,17 @@ def __validate_expense_group(expense_group: ExpenseGroup, general_settings: Work
                     'message': 'Tax Group Mapping not found'
                 })
 
+                if tax_group:
+                    Error.objects.update_or_create(
+                    workspace_id=expense_group.workspace_id,
+                    expense_attribute=category_attribute,
```
Sravanksk, line 580: why is this `category_attribute` ?

Sravanksk: should be `tax_group`
` expense_attribute= tax_group,`

Sravanksk (approved):

apps/quickbooks_online/tasks.py
Shwetabhk: yes yes fixing

PR #268, opened 2022-06-02
set qbo_realm_id to null on qbo disconnection

Sravanksk:


apps/workspaces/signals.py
```diff
@@ -34,6 +34,8 @@ def post_delete_qbo_connection(workspace_id):
     """
     workspace = Workspace.objects.get(id=workspace_id)
     if workspace.onboarding_state in ('CONNECTION', 'MAP_EMPLOYEES', 'EXPORT_SETTINGS'):
+        EmployeeMapping.objects.filter(workspace_id=workspace_id).delete()
```
ashwin1111, line 37: At this point, there won't be any mappings, this is not needed

ashwin1111 (commented):
One more thing is missed, where we check the realm_id match for pre and post save for connected orgs

ashwin1111 (approved):

PR #274, opened 2022-06-06
Allow onboarding orgs to connect to a new account 

Sravanksk:


apps/workspaces/views.py
```diff
@@ -305,16 +306,24 @@ def post(self, request, **kwargs):
...
                 qbo_connector = QBOConnector(qbo_credentials, workspace_id=kwargs['workspace_id'])
                 company_info = qbo_connector.get_company_info()
                 qbo_credentials.company_name = company_info['CompanyName']
-                qbo_credentials.save()
+                qbo_credentials.country = company_info['Country']
+                if realm_id == workspace.qbo_realm_id:
+                    qbo_credentials.save()
+                else:
+                    assert_valid(realm_id == workspace.qbo_realm_id,
+                                 'Please choose the correct Quickbooks online account')
```
Sravanksk, line 323: ^^ This check will handle the following case - 
`onboarding completed orgs won't be able to connect to a different QBO account`

ashwin1111 (approved):

PR #312, opened 2022-08-29
adding active flag for department and class

labhvam5:


apps/quickbooks_online/utils.py
```diff
@@ -252,7 +252,8 @@ def sync_departments(self):
                 'attribute_type': 'DEPARTMENT',
                 'display_name': 'Department',
                 'value': department['FullyQualifiedName'],
-                'destination_id': department['Id']
+                'destination_id': department['Id'],
+                'active' : True
```
Sravanksk, line 256: Let's set this active flag with the value we get from QBO response. 
Please check the QBO API response

apps/mappings/tasks.py
```diff
@@ -334,7 +334,7 @@ def create_fyle_categories_payload(categories: List[DestinationAttribute], works
                 payload.append({
                     'name': category.value,
                     'code': category.destination_id,
-                    'is_enabled': True if category.active is None else category.active,
```
Sravanksk, line 337: make sure you don't have any destination_attributes with active = null, else this would break

apps/quickbooks_online/utils.py
```diff
@@ -4,6 +4,7 @@
 from datetime import datetime, timedelta
 import logging
 import json
+from xml.etree.ElementTree import QName
```
Sravanksk, line 7: remove this unused import

Sravanksk (approved):

PR #321, opened 2022-09-09
Update last export detail sync'ly during last export

ashwin1111:


apps/quickbooks_online/tasks.py
```diff
@@ -619,14 +620,17 @@ def schedule_cheques_creation(workspace_id: int, expense_group_ids: List[str]):
                 task_log.status = 'ENQUEUED'
                 task_log.save()
 
-            chain.append('apps.quickbooks_online.tasks.create_cheque', expense_group, task_log.id)
+            last_export = False
```
ashwin1111, line 623: if I have a green for this POC, will be adding to other export types

Shwetabhk (approved):

PR #345, opened 2022-10-25
validate if vendor mapped is active

harshithaputtaswamy:


apps/quickbooks_online/tasks.py
```diff
@@ -538,7 +538,16 @@ def __validate_expense_group(expense_group: ExpenseGroup, general_settings: Work
...
                 entity = entity.destination_employee
             else:
-                entity = entity.destination_vendor
+                if not entity.destination_vendor.active:
+                    bulk_errors.append({
+                        'row': None,
+                        'expense_group_id': expense_group.id,
+                        'value': expense_group.description.get('employee_email'),
+                        'type': 'Employee Mapping',
+                        'message': 'Employee is mappend to inactive vendor on QuickBooks'
```
ashwin1111, line 547: in UI we don't show the mapped row. So this would be misleading. We can throw the default error message.

apps/quickbooks_online/tasks.py
```diff
@@ -538,7 +538,16 @@ def __validate_expense_group(expense_group: ExpenseGroup, general_settings: Work
             if general_settings.employee_field_mapping == 'EMPLOYEE':
                 entity = entity.destination_employee
             else:
-                entity = entity.destination_vendor
```
ashwin1111, line 541: entity = entity.destination_vendor if entity.destination_vendor.active else None

ashwin1111 (changes requested):

ashwin1111 (approved):

PR #365, opened 2023-01-12
Wrong number of imports issue fix

vishalpandeyvip:
I checked with only reimbursable and only ccc expenses, in that case, the query works fine, but fails when we have both reimbursable and CCC exxpenses in same report.
IMO the issue was because of the variable initialization, some of the variables `settled_at, approved_at, last_paid_at` were initialized in the reimbursable block but not set to None for ccc block and hence these values got carried in ccc's query leading to a wrong number of imports.


apps/fyle/tasks.py
```diff
@@ -108,7 +108,7 @@ def async_create_expense_groups(workspace_id: int, fund_source: List[str], task_
                 if expense_group_settings.expense_state == 'PAYMENT_PROCESSING':
                     settled_at = last_synced_at
 
-                if expense_group_settings.expense_state == 'PAID':
+                elif expense_group_settings.expense_state == 'PAID':
```
NileshPant1999, line 111: why are we changing this to elif 😕

NileshPant1999: i mean what is the point of changing this, expense_state will either be `PAYMENT_PROCESSING` or `PAID`

PR #367, opened 2023-01-12
only ccc export setting save bug fix

vishalpandeyvip:


apps/workspaces/apis/export_settings/serializers.py
```diff
@@ -38,6 +38,7 @@ class ExpenseGroupSettingsSerializer(serializers.ModelSerializer):
     reimbursable_export_date_type = serializers.CharField(allow_null=True, allow_blank=True, required=False)
     corporate_credit_card_expense_group_fields = serializers.ListField(allow_null=True, required=False)
     ccc_export_date_type = serializers.CharField(allow_null=True, allow_blank=True, required=False)
+    ccc_expense_state = serializers.CharField(allow_null=True, allow_blank=True, required=False)
```
labhvam5, line 41: Add the same line for `expense_state` as well . 
```suggestion
    expense_state = serializers.CharField(allow_null=True, allow_blank=True, required=False)
```

labhvam5 (approved):

PR #393, opened 2023-03-02
change in views for ExpenseView

anishfyle:


apps/fyle/views.py
```diff
@@ -474,7 +474,7 @@ def get_queryset(self):
         }
         is_skipped = self.request.query_params.get('is_skipped')
         if is_skipped:
```
ashwin1111, line 476: if is_skipped == 'true':

ashwin1111 (approved):

PR #407, opened 2023-04-13
Duplicate vendor name exists

ruuushhh:


apps/quickbooks_online/utils.py
```diff
@@ -81,8 +81,10 @@ def get_or_create_vendor(self, vendor_name: str, email: str = None, create: bool
         vendor_name = vendor_name.replace('&', '%26')  # Replace '&' with %26
 
         vendor = self.connection.vendors.search_vendor_by_display_name(vendor_name)
+        employee = DestinationAttribute.objects.filter(display_name='employee', value=vendor_name, workspace_id=self.workspace_id).count()
```
ashwin1111, line 84: let's filter on `attribute_type` to maintain consistency. Also, let's do in a single query instead of 2. `attribute_type` is employee / vendor and rest filters

ashwin1111 (approved):

PR #441, opened 2023-06-05
Handle export exceptions

ruuushhh:


apps/quickbooks_online/exceptions.py
```diff
@@ -0,0 +1,124 @@
...
+from apps.tasks.models import TaskLog, Error
+from apps.fyle.models import ExpenseGroup
+from qbosdk.exceptions import WrongParamsError, InvalidTokenError
+
+
+logger = logging.getLogger(__name__)
+logger.level = logging.INFO
+
+
+def handle_quickbooks_error(exception, expense_group: ExpenseGroup, task_log: TaskLog, export_type: str):
```
ashwin1111, line 17: hope nothing is changed in this function, pls reply incase if there are any changes

apps/quickbooks_online/exceptions.py
```diff
@@ -0,0 +1,124 @@
...
+def handle_export_exceptions(task_name):
+    def decorator(func):
+        def new_fn(expense_group: ExpenseGroup, task_log_id: int, *args):
+            error = {
+                'task': task_name,
+                'workspace_id': expense_group.workspace_id,
+                'alert': False,
+                'message': None,
+                'response': None
+            }
```
ashwin1111, line 71: I don't think this is needed at all

apps/quickbooks_online/tasks.py
```diff
@@ -808,6 +693,7 @@ def schedule_qbo_expense_creation(workspace_id: int, expense_group_ids: List[str
             chain.run()
 
 
+@handle_export_exceptions(task_name="Create QBO Expense")
```
ashwin1111, line 696: I hope you would've removed try and moved all logic 1 block outside, let me know if you've changed any other things

apps/quickbooks_online/exceptions.py
ruuushhh: I just moved that function in exceptions file everything else is same

apps/quickbooks_online/exceptions.py
ruuushhh: We will need this in default exception right?
`except Exception:
                error = traceback.format_exc()
                task_log.detail = {
                    'error': error
                }
                task_log.status = 'FATAL'
                task_log.save()
                logger.error('Something unexpected happened workspace_id: %s %s', 
                             task_log.workspace_id, task_log.detail)`

apps/quickbooks_online/tasks.py
```diff
@@ -24,6 +24,7 @@
 from .models import Bill, BillLineitem, Cheque, ChequeLineitem, CreditCardPurchase, CreditCardPurchaseLineitem, \
     JournalEntry, JournalEntryLineitem, BillPayment, BillPaymentLineitem, QBOExpense, QBOExpenseLineitem
 from .utils import QBOConnector
+from .exceptions import handle_export_exceptions, handle_quickbooks_error
```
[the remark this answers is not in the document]
ruuushhh: for `create_bill_payment` function as it takes `workspace_id` as default argument so we won't be able to use decorator for that

apps/quickbooks_online/exceptions.py
```diff
@@ -0,0 +1,124 @@
...
+                        'is_resolved': False
+                    }
+                )
+
+    task_log.status = 'FAILED'
+    task_log.detail = None
+    task_log.quickbooks_errors = errors
+    task_log.save()
+
+def handle_export_exceptions(task_name):
```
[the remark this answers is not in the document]
ruuushhh: if we are using error then we might need this right?
https://github.com/fylein/fyle-qbo-api/pull/441#discussion_r1219640153

apps/quickbooks_online/tasks.py
ruuushhh: yes will check once again

apps/quickbooks_online/exceptions.py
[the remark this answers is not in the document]
ashwin1111: replied

apps/quickbooks_online/tasks.py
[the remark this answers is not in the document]
ashwin1111: Screenshot 2023-06-08 at 4 15 08 PM
duplicate 23-51, move it to a separate function. Create 1 more decorator for handling bill payment exception. Expect args according to what you need, then re use the same function that we separated out above.

apps/quickbooks_online/exceptions.py
```diff
@@ -0,0 +1,114 @@
...
+                        'is_resolved': False
+                    }
+                )
+
+    task_log.status = 'FAILED'
+    task_log.detail = None
+    task_log.quickbooks_errors = errors
+    task_log.save()
+
+def handle_export_exceptions():
```
ashwin1111, line 62: handle bill payment too

apps/quickbooks_online/exceptions.py
```diff
@@ -59,12 +59,18 @@ def handle_quickbooks_error(exception, expense_group: ExpenseGroup, task_log: Ta
...
-def handle_export_exceptions():
+def handle_export_exceptions(bill_payment=False):
     def decorator(func):
-        def new_fn(expense_group: ExpenseGroup, task_log_id: int, *args):
-            task_log = TaskLog.objects.get(id=task_log_id)
+        def new_fn(*args):
+            if not bill_payment:
+                task_log_id = args[1]
+                task_log = TaskLog.objects.get(id=task_log_id)
+                expense_group = args[0]
```
ashwin1111, line 68: add this in line 66 for better readability

apps/quickbooks_online/exceptions.py
```diff
@@ -59,12 +59,18 @@ def handle_quickbooks_error(exception, expense_group: ExpenseGroup, task_log: Ta
     task_log.quickbooks_errors = errors
     task_log.save()
 
-def handle_export_exceptions():
+def handle_export_exceptions(bill_payment=False):
     def decorator(func):
-        def new_fn(expense_group: ExpenseGroup, task_log_id: int, *args):
-            task_log = TaskLog.objects.get(id=task_log_id)
+        def new_fn(*args):
+            if not bill_payment:
```
ashwin1111, line 65: can we handle `last_export` as well?

apps/quickbooks_online/exceptions.py
```diff
@@ -59,12 +59,18 @@ def handle_quickbooks_error(exception, expense_group: ExpenseGroup, task_log: Ta
...
+                task_log_id = args[1]
+                task_log = TaskLog.objects.get(id=task_log_id)
+                expense_group = args[0]
+            else:
+                expense_group = args[0].expense_group
```
ashwin1111, line 70: `def process_bill_payments(bill: Bill, workspace_id: int, task_log: TaskLog):    `
1st argument expects bill. 2nd arg workspace id is not added in else case. Check 3rd arg as well

apps/quickbooks_online/tasks.py
```diff
@@ -945,6 +945,40 @@ def check_expenses_reimbursement_status(expenses):
 
     return all_expenses_paid
 
+@handle_export_exceptions(bill_payment=True)
```
ashwin1111, line 948: let's name the decorator handle_qbo_exceptions if we are able to bring all handing under this single decorator

ashwin1111 (approved):

PR #446, opened 2023-06-20
Refactor complete for quickbooks, tasks, users

ruuushhh:


apps/quickbooks_online/actions.py
```diff
@@ -0,0 +1,93 @@
...
+            status=status.HTTP_200_OK
+        )
+    except QBOCredential.DoesNotExist:
+        return Response(
+            data={
+                'message': 'QBO credentials not found in workspace'
+            },
+            status=status.HTTP_400_BAD_REQUEST
+        )
+    except WrongParamsError:
```
ashwin1111, line 32: Club InvalidTokenError, both does the same job. Adjust the response message in a slightly generic way to cover these 2 exceptions

apps/quickbooks_online/actions.py
```diff
@@ -0,0 +1,93 @@
...
+    if workspace.destination_synced_at:
+        time_interval = datetime.now(timezone.utc) - workspace.destination_synced_at
+
+    if workspace.destination_synced_at is None or time_interval.days > 0:
+        quickbooks_credentials = QBOCredential.get_active_qbo_credentials(workspace_id)
+        quickbooks_connector = QBOConnector(quickbooks_credentials, workspace_id=workspace_id)
+        quickbooks_connector.sync_dimensions()
+
+        workspace.destination_synced_at = datetime.now()
+        workspace.save(update_fields=['destination_synced_at'])
```
ashwin1111, line 93: new line at end of file

apps/quickbooks_online/views.py
```diff
@@ -27,32 +19,23 @@ class VendorView(generics.ListCreateAPIView):
     """
     Vendor view
     """
+    queryset = DestinationAttribute.objects.all()[:10]
```
ashwin1111, line 22: is slicing working properly? noticed a message on slack around this area, pls confirm

apps/quickbooks_online/views.py
```diff
@@ -27,32 +19,23 @@ class VendorView(generics.ListCreateAPIView):
...
-    def get_queryset(self):
-        search_term = self.request.query_params.get('search_term')
-        if search_term:
-            return DestinationAttribute.objects.filter(
-                attribute_type='VENDOR', active=True, workspace_id=self.kwargs['workspace_id'],value__icontains=search_term).order_by('value')[:10]
-        return DestinationAttribute.objects.filter(
-            attribute_type='VENDOR', active=True, workspace_id=self.kwargs['workspace_id']).order_by('value')[:10]
-
+    filter_backends = (DjangoFilterBackend,)
+    filterset_fields = {'value': {'icontains'}, 'attribute_type': {'exact', 'in'}, 'active':{'exact'}, 'workspace_id': {'exact'}}
```
ashwin1111, line 26: `'active': {'exact'}` one space after colon. Also, should we send workspace_id explicitly from client as query param or does it pick from the URL?

apps/quickbooks_online/views.py
```diff
@@ -27,32 +19,23 @@ class VendorView(generics.ListCreateAPIView):
...
+    ordering_fields = ['value']
 
 class EmployeeView(generics.ListCreateAPIView):
     """
     Employee view
     """
+    queryset = DestinationAttribute.objects.all()
```
ashwin1111, line 33: why there is no limit to this API?

apps/quickbooks_online/views.py
```diff
@@ -61,47 +44,8 @@ class PreferencesView(generics.RetrieveAPIView):
...
-                data={
-                    'message': 'Invalid token, try to refresh it'
-                },
-                status=status.HTTP_400_BAD_REQUEST
-            )
-
+        return get_preferences(kwargs['workspace_id'])
+       
 
 class BillPaymentView(generics.CreateAPIView):
```
ashwin1111, line 50: this is a dead endpoint, pls remove

apps/quickbooks_online/views.py
```diff
@@ -186,67 +98,32 @@ class DestinationAttributesView(generics.ListAPIView):
...
-            'active': True
-        }
-
-        if display_name:
-            display_name = display_name.split(',')
-            filters['display_name__in'] = display_name
-
-        return DestinationAttribute.objects.filter(**filters).order_by('value')
-
+#slice problem
```
ashwin1111, line 108: todo

apps/quickbooks_online/views.py
```diff
@@ -186,67 +98,32 @@ class DestinationAttributesView(generics.ListAPIView):
...
-        if search_term:
-            filters['value__icontains'] = search_term
-
-        if active and active.lower() == 'true':
-            filters['active'] = True
-
-        return DestinationAttribute.objects.filter(**filters).order_by('value')[:30]
-
+    filter_backends = (DjangoFilterBackend,)
+    filterset_fields = {'workspace_id': {'exact'}, 'attribute_type': {'exact', 'in'}, 'display_name': {'exact', 'in'}, 'active': {'exact'}}
```
ashwin1111, line 117: why are we not adding value? the existing view would filter by value as well

apps/tasks/urls.py
```diff
@@ -18,6 +18,6 @@
 from .views import TasksView, TasksByIdView
 
 urlpatterns = [
-    path('', TasksByIdView.as_view()),
+    path('<int:id>/', TasksByIdView.as_view()),
```
ashwin1111, line 21: these are dead endpoints, let's remove, please verify once

apps/quickbooks_online/views.py
ruuushhh: no we will have to send explicitly

ashwin1111: we should find a way to pick it from url

ashwin1111 (approved):

PR #447, opened 2023-06-21
Refactor done workspaces

ruuushhh:


apps/workspaces/actions.py
```diff
@@ -0,0 +1,117 @@
...
+from apps.fyle.models import ExpenseGroupSettings
+from .utils import generate_qbo_refresh_token, assert_valid
+from .serializers import WorkspaceSerializer, QBOCredentialSerializer, \
+    WorkSpaceGeneralSettingsSerializer, LastExportDetailSerializer
+from .signals import post_delete_qbo_connection
+
+
+User = get_user_model()
+
+def qbo_workspace(user, org_id, org_name, org_currency):
```
ashwin1111, line 24: update_or_create_workspace

apps/workspaces/views.py
```diff
@@ -157,65 +136,19 @@ def get(self, request):
...
-
-            if workspace.onboarding_state == 'CONNECTION':
-                workspace.onboarding_state = 'MAP_EMPLOYEES'
-            workspace.save()
-
-            # Return the QBO credentials as serialized data
-            return Response(data=QBOCredentialSerializer(qbo_credentials).data, status=status.HTTP_200_OK)
-
+            return connect_qbo_oauth(refresh_token, realm_id, redirect_uri)
         except qbo_exc.UnauthorizedClientError:
```
ashwin1111, line 152: let's club all below exception and return a generic message

ashwin1111: make sure you do logger.info()

apps/workspaces/views.py
```diff
@@ -328,59 +222,30 @@ def post(self, request, *args, **kwargs):
...
 
-class LastExportDetailView(viewsets.ViewSet):
+class LastExportDetailView(generics.RetrieveAPIView):
     """
     Last Export Details
     """
+    lookup_field = 'workspace_id'
+    lookup_field_kwarg = 'workspace_id'
+
+    quryset = LastExportDetail.objects.first()
```
ashwin1111, line 232: typo

ashwin1111: is it possible to add last_exported_at__isnull=False and total_expense_groups_count > 0?

apps/workspaces/views.py
```diff
@@ -328,59 +222,30 @@ def post(self, request, *args, **kwargs):
...
-                'email': admin.email
-            })
-
         return Response(
                 data=admin_email,
                 status=status.HTTP_200_OK
             )
 
-class SetupE2ETestView(viewsets.ViewSet):
+class SetupE2ETestView(generics.CreateAPIView):
```
ashwin1111, line 248: can you move whole logic to actions?

apps/workspaces/views.py
ruuushhh: not with RetrieveAPIView

ashwin1111: Screenshot 2023-06-26 at 5 04 45 PM
try this out @ruuushhh

apps/workspaces/views.py
```diff
@@ -56,39 +42,14 @@ def post(self, request):
         """
         Create a Workspace
         """
-
         access_token = request.META.get('HTTP_AUTHORIZATION')
+        user = request.user
         fyle_user = get_fyle_admin(access_token.split(' ')[1], None)
```
ashwin1111, line 47: move all these logic to actions, there shouldn't be anything happening in views

ruuushhh: tried with this only @ashwin1111 but that didn't work
So had to get `request.META.get('HTTP_AUTHORIZATION')` from outside

apps/workspaces/views.py
```diff
@@ -157,100 +102,24 @@ def get(self, request):
...
-            return Response({'message': 'QBO Application not found'}, status=status.HTTP_404_NOT_FOUND)
-
-        except qbo_exc.WrongParamsError as e:
-            return Response(json.loads(e.response), status=status.HTTP_400_BAD_REQUEST)
-
-        except qbo_exc.InternalServerError:
-            return Response({'message': 'Wrong/Expired Authorization code'}, status=status.HTTP_401_UNAUTHORIZED)
-
-
-    def patch(self, request, **kwargs):
```
ashwin1111, line 232: we'll need patch method to do post_delete_qbo_connection and revoke_refresh_token. Can do something similar to this, and then call the couple of methods
Screenshot 2023-07-06 at 12 30 04 PM

ashwin1111 (approved):

PR #448, opened 2023-06-21
post-date-integration-support for qbo changes

DhaaraniCIT:


tests/test_quickbooks_online/test_utils.py
```diff
@@ -303,9 +303,6 @@ def test_construct_cheque_item_and_account_based(create_cheque_item_and_account_
     cheque,cheque_lineitems = create_cheque_item_and_account_based
     cheque_object = qbo_connection._QBOConnector__construct_cheque(cheque=cheque,cheque_lineitems=cheque_lineitems)
 
-    cheque_object['Line'][0]['DetailType'] == 'ItemBasedExpenseLineDetail'
-    cheque_object['Line'][1]['DetailType'] == 'AccountBasedExpenseLineDetail'
```
ashwin1111, line 307: why is this removed?

tests/test_quickbooks_online/fixtures.py
```diff
@@ -409,44 +409,44 @@
...
             {
-                "Description":"user9@fyleforgotham.in - Concrete - 2023-04-19 - C/2023/04/R/13 -  - None/app/main/#/enterprise/view_expense/txT4kpKidaAdLm?org_id=or79Cob97KSh",
-                "DetailType":"ItemBasedExpenseLineDetail",
+                "Description":"user9@fyleforgotham.in - Food - 2023-04-19 - C/2023/04/R/13 -  - None/app/main/#/enterprise/view_expense/txT4kpMbiadw?org_id=or79Cob97KSh",
+                "DetailType":"AccountBasedExpenseLineDetail",
                 "Amount":1.0,
-                "ItemBasedExpenseLineDetail":{
-                    "ItemRef":{
-                    "value":"3"
+                "AccountBasedExpenseLineDetail":{
```
ashwin1111, line 415: why are these changed?

tests/test_quickbooks_online/test_utils.py
DhaaraniCIT: codecov gave command that 'there is no effect', so changed

tests/test_quickbooks_online/fixtures.py
DhaaraniCIT: the unit test was failing, because the positions were differing when compared to the fixture the and resulted data

ashwin1111: ideally your changes doesn't touch these areas no? please discuss the failure with @labhvam5

Shwetabhk: I don't think the order matters, the failure could've been for some other reason, please check it with @labhvam5

Shwetabhk: Ideally there should be no change in this file

tests/test_quickbooks_online/test_utils.py
Shwetabhk: I think these are supposed to be asserts

Shwetabhk:
```python
    assert cheque_object['Line'][0]['DetailType'] == 'ItemBasedExpenseLineDetail'
    assert cheque_object['Line'][1]['DetailType'] == 'AccountBasedExpenseLineDetail'
```

Shwetabhk (changes requested):

tests/test_fyle/test_models.py
```diff
@@ -99,3 +102,44 @@ def test_format_date():
     date_string = _format_date('2022-05-13T09:32:06.643941Z')
 
     assert date_string == parser.parse('2022-05-13T09:32:06.643941Z')
+
+def test_support_post_date_integrations(mocker, db, api_client, test_connection):
```
Shwetabhk, line 106: test name is not right, there is no where in this test where we're asserting whether the posted_at got save correctly in the db or not
Please add an assert for thay

tests/test_fyle/test_models.py
```diff
@@ -99,3 +102,44 @@ def test_format_date():
...
+
+    assert task_log.status == 'COMPLETE'
+
+	#Export assert
+    mocker.patch(
+        'apps.workspaces.views.export_to_qbo',
+        return_value=None
+    )
+
+    workspace_id = 3
```
Shwetabhk, line 140: this test is unnecessary and is already tested in views.py
We shouldn't add tests for views in test_models.py file

tests/test_fyle/test_models.py
```diff
@@ -99,3 +102,44 @@ def test_format_date():
...
+    expense_group_settings = ExpenseGroupSettings.objects.get(workspace_id=workspace_id)
+    expense_group_settings.reimbursable_export_date_type = 'last_spent_at'
+    expense_group_settings.ccc_export_date_type = 'posted_at'
+    expense_group_settings.save()
+
+    create_expense_groups(workspace_id, ['PERSONAL', 'CCC'], task_log)
+
+    task_log = TaskLog.objects.get(id=task_log.id)
+
+    assert task_log.status == 'COMPLETE'
```
Shwetabhk, line 132: Add more assertions to ensure that posted_at is saved correctly and is same as what's there in the fixtures

tests/test_fyle/test_models.py
Shwetabhk: if you want to add a unit test for the whole flow add the test in test_quickbooks_online/test_models.py

Triggering the view cannot make sure that the correct posted_at is added because tests don't have access to DjangoQ.

tests/test_quickbooks_online/fixtures.py
DhaaraniCIT: Got green for the changes from @labhvam5

tests/test_quickbooks_online/test_utils.py
```diff
@@ -303,9 +303,9 @@ def test_construct_cheque_item_and_account_based(create_cheque_item_and_account_
...
-
+    assert cheque_object['Line'][1]['DetailType'] == 'ItemBasedExpenseLineDetail'
+    assert cheque_object['Line'][0]['DetailType'] == 'AccountBasedExpenseLineDetail'
+
```
Shwetabhk, line 308: remove extra spaces

Shwetabhk (approved):
@DhaaraniCIT approved with 1 minor comment, please fix it off before merging

PR #453, opened 2023-07-06
Last export p0 fixed

ruuushhh:


apps/quickbooks_online/exceptions.py
```diff
@@ -114,6 +114,8 @@ def new_fn(*args):
                 logger.error('Something unexpected happened workspace_id: %s %s', 
                              task_log.workspace_id, task_log.detail)
 
+            if args[2] == True and args[2] > 0 and not bill_payment:
```
ashwin1111, line 117: add `if len(args) > 2` at the start. args[2] > 0 is not required

ashwin1111 (approved):

NileshPant1999 (approved):

PR #455, opened 2023-07-07
Introduce flake8 for linting

ruuushhh:


.flake8
```diff
@@ -0,0 +1,23 @@
...
+    funding_sources_module/*,
+    libs/*,
+    migrations/*,
+    models/*,
+    payouts_module/*,
+    scripts/*,
+    test_reports/*,
+    tests/*,
+    webhooks_module/*,
+    gunicorn_config.py
```
ashwin1111: let's not exclude any of these for now

.flake8
```diff
@@ -0,0 +1,23 @@
+[flake8]
+extend-ignore = E712, E711
```
ashwin1111, line 2: would be good to add a comment of what these are

.github/workflows/lint_and_format_check.yml
```diff
@@ -0,0 +1,25 @@
+name: Lint and Format Check
```
ashwin1111, line 1: this is not needed since we run this on dockerfile

.pre-commit-config.yaml
```diff
@@ -0,0 +1,40 @@
+default_stages: [commit]
+fail_fast: true
+default_language_version:
+  python: python3
+exclude: 'venv|.venv|.git|README.md|beneficial_owners_module|customers_module|dev|dwolla_service|funding_sources_module|libs|migrations|models|payouts_module|scripts|test_reports|tests|webhooks_module|gunicorn_config.py'
```
ashwin1111: remove all these

ashwin1111 (approved):

.flake8
```diff
@@ -3,7 +3,8 @@ extend-ignore =
     # Comparison to true should be 'if cond is true:' or 'if cond:'
     E712,
     # Comparison to None should be 'cond is None:' (E711)
-    E711
+    E711,
+    W503
```
ashwin1111, line 7: add comments

ashwin1111 (approved):

Shwetabhk (approved):

PR #471, opened 2023-08-07
bug fix: use filter instead of get query

NileshPant1999:


ruuushhh (approved):

apps/workspaces/actions.py
```diff
@@ -105,9 +105,10 @@ def get_workspace_admin(workspace_id: int):
     users = workspace.user.all()
     for user in users:
         admin = User.objects.get(user_id=user)
-        name = ExpenseAttribute.objects.get(value=admin.email, workspace_id=workspace_id, attribute_type='EMPLOYEE').detail['full_name']
+        employee = ExpenseAttribute.objects.filter(value=admin.email, workspace_id=workspace_id, attribute_type='EMPLOYEE').first()
+        if employee:
+            admin_email.append({'name': employee.detail['name'], 'email': admin.email})
```
ruuushhh, line 110: `'name': employee.detail['full_name']`
it should be this

PR #476, opened 2023-09-04
Update skipped expenses in Fyle

ashwin1111:


apps/fyle/queue.py
```diff
@@ -0,0 +1,5 @@
+from django_q.tasks import async_task
+
+
+def async_post_accounting_export_summary(org_id: str, workspace_id: int) -> None:
```
Shwetabhk, line 4: Add some comments to the function please along with description for function params

apps/fyle/tasks.py
```diff
@@ -48,6 +52,39 @@ def create_expense_groups(workspace_id: int, fund_source: List[str], task_log: T
     return task_log
 
 
+def __get_updated_accounting_export_summary(expense_id: str, is_synced: bool) -> dict:
+    return {
+        'id': expense_id,
+        'state': 'SKIPPED',
```
Shwetabhk, line 58: lets also pass state from Functions Param no, in case we want to use the same function for any other state but Skipped

ashwin1111: yeah, that will be covered in the next PR

apps/fyle/tasks.py
```diff
@@ -48,6 +52,39 @@ def create_expense_groups(workspace_id: int, fund_source: List[str], task_log: T
...
+    expense_to_be_updated = []
+    expenses_to_be_skipped = Expense.objects.filter(
+        final_query,
+        id__in=expenses_object_ids,
+        expensegroup__isnull=True,
+        org_id=workspace.fyle_org_id
+    )
+
+    for expense in expenses_to_be_skipped:
+        updated_accounting_export_summary = __get_updated_accounting_export_summary(expense.expense_id, False)
```
Shwetabhk, line 75: Need not store it in a vairable, can directly call the function on line 80

apps/fyle/tasks.py
```diff
@@ -48,6 +52,39 @@ def create_expense_groups(workspace_id: int, fund_source: List[str], task_log: T
...
+        expense_to_be_updated.append(
+            Expense(
+                id=expense.id,
+                is_skipped=True,
+                accounting_export_summary=updated_accounting_export_summary
```
Shwetabhk, line 80:
```suggestion
                accounting_export_summary=__get_updated_accounting_export_summary(expense.expense_id, 'SKIPPED', False)
```

Shwetabhk (approved):

PR #478, opened 2023-09-04
Post failed expense

ashwin1111:


apps/fyle/helpers.py
```diff
@@ -261,3 +261,26 @@ def update_expenses_in_progress(in_progress_expenses: List[Expense]) -> None:
...
+    :param in_progress_expenses: In progress expenses
+    """
+    expense_to_be_updated = []
+    for expense in in_progress_expenses:
+        expense_to_be_updated.append(
+            Expense(
+                id=expense.id,
+                accounting_export_summary=get_updated_accounting_export_summary(
+                    expense.expense_id,
+                    'ERROR',
```
Shwetabhk, line 278: Something for later, we should use enums for all the strings we use in backend

apps/workspaces/signals.py
```diff
@@ -35,3 +36,13 @@ def post_delete_qbo_connection(workspace_id):
         workspace.onboarding_state = 'CONNECTION'
         workspace.qbo_realm_id = None
         workspace.save()
+
+
+@receiver(post_save, sender=LastExportDetail)
+def run_post_save_last_export_detail(sender, instance: LastExportDetail, **kwargs):
```
Shwetabhk, line 42: Any advantage of adding this here and not at the end of the chain

Shwetabhk (approved):

PR #480, opened 2023-09-05
Add workspace_id to Expenses and fill data

ashwin1111:


ashwin1111:
Filled data on staging already for both. On prod, there are no expense without org_id. Once this reach prod, will run workspace id generation script.

sql/scripts/020-fill-org-id-expenses.sql
```diff
@@ -0,0 +1,14 @@
+rollback;
```
Shwetabhk, line 1: check if there are any expenses which do not have expense_groups
Skipped expenses would also not have expenses_groups

sql/scripts/020-fill-org-id-expenses.sql
```diff
@@ -0,0 +1,14 @@
...
+begin;
+
+with expense_groups as (
```
Shwetabhk, line 4: check if there are any expenses with no expense_groups

Shwetabhk (approved):

PR #501, opened 2023-10-13
Fix pagination and post summary grouped by fund source

ashwin1111:


apps/fyle/tasks.py
```diff
@@ -166,26 +165,32 @@ def sync_dimensions(fyle_credentials):
         logger.info('Invalid Token for fyle')
 
 
-def post_accounting_export_summary(org_id: str, workspace_id: int) -> None:
+def post_accounting_export_summary(org_id: str, workspace_id: int, fund_source: str=None) -> None:
```
Shwetabhk, line 168:
```suggestion
def post_accounting_export_summary(org_id: str, workspace_id: int, fund_source: str = None) -> None:
```

Shwetabhk (approved):

PR #519, opened 2023-11-14
Import projects base imp

labhvam5:


apps/mappings/exceptions.py
```diff
@@ -50,3 +51,60 @@ def new_fn(workspace_id: int, *args):
...
+        }
+        try:
+            return func(expense_attribute_instance, *args)
+        except WrongParamsError as exception:
+            error['message'] = exception.message
+            error['response'] = exception.response
+            error['alert'] = True
+            import_log.status = 'FAILED'
+
+        except (QBOCredential.DoesNotExist):
```
ashwin1111, line 75: can merge QBOCredential.DoesNotExist, QBOWrongParamsError, QBOInvalidTokenError together

apps/mappings/queues.py
```diff
@@ -107,26 +109,25 @@ def construct_tasks_and_chain_import_fields_to_fyle(workspace_id):
     mapping_settings = MappingSetting.objects.filter(workspace_id=workspace_id, import_to_fyle=True)
     credentials = QBOCredential.objects.get(workspace_id=workspace_id)
 
-    task_settings = {
+    task_settings: TasksSettings = {
```
ashwin1111, line 112: `TaskSetting`, let's make dataclasses should singular, the variable can be plural - that's okay

apps/mappings/queues.py
```diff
@@ -107,26 +109,25 @@ def construct_tasks_and_chain_import_fields_to_fyle(workspace_id):
...
-                'destination_field': mapping_setting.destination_field,
-                'destination_sync_method': SYNC_METHODS[mapping_setting.destination_field],
-                'is_auto_sync_enabled': get_auto_sync_permission(mapping_setting),
-                'is_custom': False,
-            })
-
-    # Make a call to the SDK
-    # chain_import_fields_to_fyle(workspace_id, task_settings)
-    pass
+    if mapping_settings:
```
ashwin1111, line 122: if is not required

apps/mappings/queues.py
```diff
@@ -107,26 +109,25 @@ def construct_tasks_and_chain_import_fields_to_fyle(workspace_id):
...
+        for mapping_setting in mapping_settings:
+            if mapping_setting.source_field in ['PROJECT']:
```
ashwin1111, line 124: I remember same check is there in imports too, if 1 place solves it, let's keep it in API.

labhvam5: this is api only. 🤣

ashwin1111: no no, remove the other one if keeping in just API should work

ashwin1111 (approved):

PR #523, opened 2023-11-26
Category-rework: scheduling

labhvam5:


apps/mappings/helpers.py
```diff
@@ -11,20 +11,20 @@ def schedule_or_delete_fyle_import_tasks(configuration: WorkspaceGeneralSettings
     :param configuration: WorkspaceGeneralSettings Instance
     :return: None
     """
-    if configuration.import_categories or configuration.import_items or configuration.import_vendors_as_merchants:
+    if configuration.import_vendors_as_merchants:
         start_datetime = datetime.now()
         Schedule.objects.update_or_create(func='apps.mappings.tasks.auto_import_and_map_fyle_fields', args='{}'.format(configuration.workspace_id), defaults={'schedule_type': Schedule.MINUTES, 'minutes': 24 * 60, 'next_run': start_datetime})
-    elif not configuration.import_categories and not configuration.import_items and not configuration.import_vendors_as_merchants:
+    elif not configuration.import_vendors_as_merchants:
```
ashwin1111, line 17: else

apps/mappings/helpers.py
```diff
@@ -11,20 +11,20 @@ def schedule_or_delete_fyle_import_tasks(configuration: WorkspaceGeneralSettings
...
 
-def get_auto_sync_permission(mapping_setting: MappingSetting):
+def get_auto_sync_permission(workspace_general_settings: WorkspaceGeneralSettings, mapping_setting: MappingSetting = None):
     """
     Get the auto sync permission
     :return: bool
     """
     is_auto_sync_status_allowed = False
-    if (mapping_setting.destination_field == 'CUSTOMER' and mapping_setting.source_field == 'PROJECT') or mapping_setting.source_field == 'CATEGORY':
+    if (mapping_setting.destination_field == 'CUSTOMER' and mapping_setting.source_field == 'PROJECT') or workspace_general_settings.import_categories:
```
ashwin1111, line 27: if mapping_setting is not send the first part of if block would fail. add mapping_setting and rest of 1st block condition

ashwin1111 (approved):

apps/mappings/helpers.py
labhvam5: We have set the ` mapping_setting: MappingSetting = None` in the function definition . So it wont fail .

fyle_integrations_imports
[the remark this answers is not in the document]
labhvam5: sure will do.

PR #532, opened 2023-12-14
Cost-Centers rework: Scheduling

labhvam5:


apps/mappings/signals.py
```diff
@@ -55,7 +54,7 @@ def run_post_mapping_settings_triggers(sender, instance: MappingSetting, **kwarg
         new_schedule_or_delete_fyle_import_tasks(workspace_general_settings, instance)
 
     if instance.source_field == 'COST_CENTER':
-        schedule_cost_centers_creation(instance.import_to_fyle, int(instance.workspace_id))
+        new_schedule_or_delete_fyle_import_tasks(workspace_general_settings, instance)
```
ashwin1111, line 57: can we update 53rd line condition to do `if instance.source_field == 'PROJECT' or instance.source_field == 'COST_CENTER'` instead of calling it twice

ashwin1111 (approved):

PR #543, opened 2023-12-20
ImportTax + Vendors : Scheduling

labhvam5:


apps/mappings/schedules.py
```diff
@@ -32,19 +33,13 @@ def schedule_or_delete_fyle_import_tasks(workspace_general_settings: WorkspaceGe
...
-            is_custom=True
+            workspace_id=workspace_general_settings.workspace_id
         ).count()
 
         # If the import fields count is 0, delete the schedule
-        if import_fields_count == 0 and custom_field_import_fields_count == 0\
-            and not workspace_general_settings.import_categories and not workspace_general_settings.import_items:
+        if import_fields_count == 0\
+            and not workspace_general_settings.import_categories and not workspace_general_settings.import_items\
+            and not workspace_general_settings.import_tax_codes and not workspace_general_settings.import_vendors_as_merchants:
```
ashwin1111, line 42: if vendor / category / merchant / item i enabled, it would've gone into if block. These conditions inside else block is not required

ashwin1111 (approved):

PR #544, opened 2023-12-20
implementation

labhvam5:


apps/mappings/schedules.py
```diff
@@ -31,7 +32,9 @@ def schedule_or_delete_fyle_import_tasks(workspace_general_settings: WorkspaceGe
         )
 
     else:
+        # Have added the TAX_GROUP in not case because the mapping_settings persists even after import_tax_codes is set to False
```
ashwin1111, line 35: can we remove tax group row in mapping settings if tax is set to false? feel like a bad data is present in db

ashwin1111 (approved):

PR #595, opened 2024-04-18
Removed attribute_type filter

Hrishabh17:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the vendor creation process by enhancing the duplicate name check logic, ensuring more accurate vendor listings.
	- Updated the logic for checking duplicate vendor names to remove a specific filter condition, enhancing the accuracy of vendor data.
	- Modified assertion values in test files for consistency and accuracy in reflecting vendor information.
	- Updated the test case for creating a journal entry to use an updated entity ID value for testing accuracy.
	- Adjusted assertion values in test files to match the correct vendor information and reflect the expected data accurately.

### DIFF
--- a/apps/quickbooks_online/utils.py
+++ b/apps/quickbooks_online/utils.py
@@ -111,8 +111,9 @@ class QBOConnector:
         if not vendor:
             if create:
                 # safe check to avoid duplicate vendor name exist error
-                if DestinationAttribute.objects.filter(value=vendor_name, workspace_id=self.workspace_id).exists():
-                    return
+                existing_vendor = DestinationAttribute.objects.filter(value=vendor_name, workspace_id=self.workspace_id).first()
+                if existing_vendor:
+                    return existing_vendor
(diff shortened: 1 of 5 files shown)


apps/quickbooks_online/utils.py
```diff
@@ -111,8 +111,9 @@ def get_or_create_vendor(self, vendor_name: str, email: str = None, create: bool
         if not vendor:
             if create:
                 # safe check to avoid duplicate vendor name exist error
-                if DestinationAttribute.objects.filter(attribute_type__in=['CUSTOMER', 'EMPLOYEE'], value=vendor_name, workspace_id=self.workspace_id).exists():
-                    return
+                existing_vendor = DestinationAttribute.objects.filter(value=vendor_name, workspace_id=self.workspace_id).first()
+                if existing_vendor:
+                    return existing_vendor
```
ashwin1111, line 116: we shouldn't return anything here. Export would error if we would've returned a non vendor value.

PR #599, opened 2024-04-26
Fix: Vendor None Error

ruuushhh:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability by adding a check to ensure a vendor exists before accessing its ID, preventing errors in scenarios where the vendor might not be present.


ashwin1111:
Hrishabh has a fix in similar area, if it failed after his merge, will look into this

PR #608, opened 2024-05-10
Support for refund for Bill module

Ashutosh619-sudo:


apps/fyle/models.py
```diff
@@ -299,6 +299,66 @@ def _group_expenses(expenses, group_fields, workspace_id):
     return expense_groups
 
 
+def filter_expense_groups(
+    expense_groups,
+    expenses: Expense,
+    expense_group_fields,
+    configuration: WorkspaceGeneralSettings,
```
ashwin1111, line 306: accept reimbursable_export_type=None, ccc_export_type=None

apps/fyle/models.py
```diff
@@ -299,6 +299,66 @@ def _group_expenses(expenses, group_fields, workspace_id):
...
+
+    filtered_expense_groups = []
+
+    for expense_group in expense_groups:
+        expense_group_expenses_ids = expense_group["expense_ids"]
+        filtered_expenses = [
+            item for item in expenses if item.id in expense_group_expenses_ids
+        ]
+        if (
+            configuration.reimbursable_expenses_object == "EXPENSE REPORT"
```
ashwin1111, line 317: `'expense_id' not in expense_group_fields and (reimbursable_export_type in ('EXPENSE', 'BILL') or ccc_export_type == 'BILL')`

apps/fyle/models.py
```diff
@@ -299,6 +299,66 @@ def _group_expenses(expenses, group_fields, workspace_id):
...
+                        )
+            else:
+                for expense in filtered_expenses:
+                    total_amount += expense.amount
+
+                if total_amount < 0:
+                    filtered_expenses = list(
+                        filter(lambda expense: expense.amount > 0, filtered_expenses)
+                    )
+        elif configuration.reimbursable_expenses_object != "JOURNAL ENTRY":
```
ashwin1111, line 348: reimbursable_export_type != JE

apps/fyle/models.py
```diff
@@ -409,7 +405,8 @@ def create_expense_groups_by_report_id_fund_source(
             reimbursable_expense_groups,
             reimbursable_expenses,
             reimbursable_expense_group_fields,
-            general_settings,
+            general_settings.reimbursable_expenses_object,
+            general_settings.corporate_credit_card_expenses_object
```
ashwin1111, line 409: need not send this

apps/fyle/models.py
```diff
@@ -435,7 +432,8 @@ def create_expense_groups_by_report_id_fund_source(
                 filtered_corporate_credit_card_expense_groups,
                 corporate_credit_card_expenses,
                 corporate_credit_card_expense_group_field,
-                general_settings,
+                general_settings.reimbursable_expenses_object,
```
ashwin1111, line 435: send as None

tests/test_fyle/test_models.py
```diff
@@ -67,28 +67,62 @@ def test_create_reimbursement(db):
...
+    expense_group = (
+        ExpenseGroup.objects.filter(workspace=workspace).order_by('-created_at').first()
+    )
+    assert expense_group.expenses.count() == 2
+
+    expenses = data['expense_refund_invalid']
+    expense_objects = Expense.create_expense_objects(expenses, 1)
+    assert len(expense_objects) == 2
+    ExpenseGroup.create_expense_groups_by_report_id_fund_source(expense_objects, 1)
+    expense_group = (
```
ashwin1111, line 120: ideally there shouldn't be expense group created for expense_refund_invalid no?

tests/test_fyle/test_models.py
```diff
@@ -67,28 +67,62 @@ def test_create_reimbursement(db):
...
+        ExpenseGroup.objects.filter(workspace=workspace).order_by('-created_at').first()
+    )
+    assert expense_group.expenses.count() == 2
+
+
```
ashwin1111, line 125: have test cases for these 
- 2 spent at fixture with spent at config and see if 2 expense group is getting created
- group by expense and assert results
- different export types

these cases helps test everything in detail

tests/test_fyle/test_models.py
Ashutosh619-sudo: yes, it didn't create expense_group for this condition, the two present are for the valid condition, otherwise there would be 4.

ashwin1111: asserts should made better. There are many other ways to locate expense groups. You can do something like `ExpenseGroup.objects.filter(expenses__amount__gte=0)` (filter can be anything based on your fixture).

tests/test_fyle/test_models.py
```diff
@@ -67,28 +67,203 @@ def test_create_reimbursement(db):
...
+    )
+    assert expense_group == None
+
+
+def test_create_expense_group_report_id_expense_report(db):
+
+    workspace = workspace = Workspace.objects.get(id=1)
+    configuration = WorkspaceGeneralSettings.objects.get(workspace=workspace)
+
+    configuration.reimbursable_expenses_object = "EXPENSE"
```
ashwin1111, line 214: we are checking with ccc expense, can we update ccc column?

tests/test_fyle/test_models.py
Ashutosh619-sudo: I have created a new test for this, this was not very clear

tests/test_fyle/test_models.py
Ashutosh619-sudo: No ashwin this is for reimbursable expense, we did change some code there as well so I wanted to test.

ashwin1111: fund_source is CCC in the fixture, it may treat it as ccc group

ashwin1111 (approved):

PR #632, opened 2024-06-24
Model changes for Split Expenses

anishfyle:


apps/fyle/models.py
```diff
@@ -199,6 +202,7 @@ class ExpenseGroupSettings(models.Model):
     reimbursable_export_date_type = models.CharField(max_length=100, default='current_date', help_text='Export Date')
     ccc_export_date_type = models.CharField(max_length=100, default='current_date', help_text='CCC Export Date')
     import_card_credits = models.BooleanField(help_text='Import Card Credits', default=False)
+    split_expense_grouping = models.CharField(max_length=100, default='SINGLE_LINE_ITEM', choices=SPLIT_EXPENSE_GROUPING, help_text='specify line items for split expenses grouping', null=True)
```
ruuushhh, line 205: if we are giving default value then remove null=true

ruuushhh, line 205: to get default value use function
This was we follow
image

ruuushhh (approved):

apps/fyle/models.py
```diff
@@ -38,6 +40,10 @@ def get_default_ccc_expense_state():
     return 'PAID'
 
 
+def get_default_split_expense_grouping():
+    return 'SINGLE_LINE_ITEM'
```
ashwin1111, line 44: default would be `MULTIPLE_LINE_ITEM`

ashwin1111 (approved):

PR #638, opened 2024-06-27
Split Expense: Grouping function

anishfyle:


apps/fyle/models.py
```diff
@@ -432,11 +432,32 @@ def create_expense_groups_by_report_id_fund_source(
...
+                corporate_credit_card_expense_group_field,
+                workspace_id,
+            )
+
+            ccc_expenses_with_bank_transaction = list(
+                filter(lambda corporate_credit_card_expenses: corporate_credit_card_expenses.bank_transaction_id, expense_objects)
+            )
+            corporate_credit_card_expense_group_field = [ccc_expense_group_field for ccc_expense_group_field in corporate_credit_card_expense_group_field if ccc_expense_group_field not in {'spent_at', 'posted_at', 'expense_id'}]
+            corporate_credit_card_expense_group_field.append('bank_transaction_id')
+            filtered_corporate_credit_card_expense_groups = _group_expenses(
```
ruuushhh, line 450: we should extend the `filtered_corporate_credit_card_expense_groups ` array here
So that array will have both with bank transaction expense groups and without bank transaction expense group

apps/fyle/models.py
```diff
@@ -432,11 +432,42 @@ def create_expense_groups_by_report_id_fund_source(
...
+                workspace_id,
+            )
+
+            ccc_expenses_with_bank_transaction = [
+                expense for expense in expense_objects
+                if expense.bank_transaction_id
+            ]
+            corporate_credit_card_expense_group_field = [
+                field for field in corporate_credit_card_expense_group_field
+                if field not in {'spent_at', 'posted_at', 'expense_id'}
```
ashwin1111, line 455: let's actually remove only expense_id and expense_number. Ignore spent at and posted at - they should be present

ashwin1111 (approved):

PR #639, opened 2024-06-27
Split Expense: purchase number generation

anishfyle:


apps/quickbooks_online/models.py
```diff
@@ -618,7 +619,7 @@ def create_credit_card_purchase(expense_group: ExpenseGroup, map_merchant_to_ven
                 'transaction_date': get_transaction_date(expense_group),
                 'private_note': private_note,
                 'currency': expense.currency,
-                'credit_card_purchase_number': expense.expense_number if map_merchant_to_vendor else '',
+                'credit_card_purchase_number': expense_group.description['bank_transaction_id'] if expense_group_settings.split_expense_grouping == 'MULTIPLE_LINE_ITEM' else expense.expense_number if map_merchant_to_vendor else '',
```
ruuushhh, line 622: if they change the export module are we going to send `split_expense_grouping ` field as None from frontend?
If not please check for export module as well

Also if code is getting complicated then move it outside this dict, should be simpler to read

ruuushhh (approved):

apps/quickbooks_online/models.py
```diff
@@ -172,6 +172,15 @@ def get_category_mapping_and_detail_type(workspace_general_settings: WorkspaceGe
     return qbo_account, 'AccountBasedExpenseLineDetail'
 
 
+def get_credit_card_purchase_number(expense_group, expense, expense_group_settings, map_merchant_to_vendor):
+    if expense_group_settings.split_expense_grouping == 'MULTIPLE_LINE_ITEM':
```
ashwin1111, line 176: and 'bank_transaction_id' in expense_group.description

apps/quickbooks_online/models.py
```diff
@@ -172,6 +172,15 @@ def get_category_mapping_and_detail_type(workspace_general_settings: WorkspaceGe
...
+        return expense_group.description['bank_transaction_id']
+    elif map_merchant_to_vendor:
```
ashwin1111, line 178: else

apps/quickbooks_online/models.py
```diff
@@ -172,6 +172,15 @@ def get_category_mapping_and_detail_type(workspace_general_settings: WorkspaceGe
...
+        return expense.expense_number
+    else:
+        return ''
```
ashwin1111, line 181: remove it off

ashwin1111 (approved):

PR #646, opened 2024-07-12
fix qbo creds inactive exports

Hrishabh17:
## Summary by CodeRabbit

- **New Features**
  - Added functionality to manage QuickBooks Online (QBO) credential expiration statuses, ensuring they are flagged and saved as expired when necessary.

- **Bug Fixes**
  - Improved `export_to_qbo` function to check for active QBO credentials, preventing unnecessary operations when credentials are not active.

- **Tests**
  - Enhanced test cases to ensure QBO credentials are correctly set to non-expired states before various operations, ensuring more reliable test scenarios.

### DIFF
--- a/apps/quickbooks_online/exceptions.py
+++ b/apps/quickbooks_online/exceptions.py
@@ -94,6 +94,7 @@ def handle_qbo_exceptions(bill_payment=False):
 
                 if qbo_credentials:
                     qbo_credentials.is_expired = True
+                    qbo_credentials.refresh_token = None
                     qbo_credentials.save()
 
                 task_log.save()
--- a/apps/quickbooks_online/exceptions.py
+++ b/apps/quickbooks_online/exceptions.py
@@ -88,6 +88,13 @@ def handle_qbo_exceptions(bill_payment=False):
                 detail = {'expense_group_id': expense_group.id, 'message': 'QBO Account not connected / token expired'}
                 task_log.status = 'FAILED'
                 task_log.detail = detail
+                qbo_credentials = QBOCredential.objects.filter(
+                    workspace_id=expense_group.workspace_id
+                ).first()
+
+                if qbo_credentials:
+                    qbo_credentials.is_expired = True
+                    qbo_credentials.save()
 
                 task_log.save()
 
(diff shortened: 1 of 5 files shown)


apps/quickbooks_online/queue.py
```diff
@@ -171,7 +189,13 @@ def schedule_credit_card_purchase_creation(workspace_id: int, expense_group_ids:
     :param fund_source: Fund source
     :return: None
     """
-    if expense_group_ids:
+    active_qbo_credentials = QBOCredential.objects.filter(
```
ashwin1111, line 192: `QBOCredential.get_active_qbo_credentials(workspace_id)`

apps/quickbooks_online/queue.py
```diff
@@ -171,7 +189,13 @@ def schedule_credit_card_purchase_creation(workspace_id: int, expense_group_ids:
...
+        workspace_id=workspace_id,
```
ashwin1111, line 193: also do this at the start of export_to_qbo() func, so that it's done only in 1 place and can avoid a lot of db queries

ashwin1111: if no active creds, then just stop execution

ashwin1111: also in handle_qbo_exceptions(), let's update qbo_creds is expired to true in (QBOCredential.DoesNotExist, InvalidTokenError) block

apps/quickbooks_online/queue.py
Hrishabh17: get_active_qbo_credentials  uses objects.get and we are not handling the QBOCredential.DoesNotExist in export_to_qbo or any caller func.

apps/quickbooks_online/exceptions.py
```diff
@@ -88,6 +88,13 @@ def new_fn(*args):
                 detail = {'expense_group_id': expense_group.id, 'message': 'QBO Account not connected / token expired'}
                 task_log.status = 'FAILED'
                 task_log.detail = detail
+                qbo_credentials = QBOCredential.objects.filter(
+                    workspace_id=expense_group.workspace_id
+                ).first()
+
+                if qbo_credentials:
+                    qbo_credentials.is_expired = True
```
ashwin1111, line 96: set refresh token to null as well

ashwin1111 (approved):

PR #717, opened 2025-01-08
fix: re structure direct export

ruuushhh:
### Description
fix: re structure direct export

## Clickup
https://app.clickup.com/

### DIFF
--- a/apps/fyle/tasks.py
+++ b/apps/fyle/tasks.py
@@ -17,7 +17,6 @@ from apps.fyle.helpers import (
     handle_import_exception,
 )
 from apps.fyle.models import Expense, ExpenseFilter, ExpenseGroup, ExpenseGroupSettings
-from apps.fyle.queue import async_post_accounting_export_summary
 from apps.tasks.models import TaskLog
 from apps.workspaces.actions import export_to_qbo
 from apps.workspaces.models import FyleCredential, Workspace, WorkspaceGeneralSettings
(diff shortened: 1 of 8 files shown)


apps/fyle/tasks.py
```diff
@@ -60,8 +59,8 @@ def group_expenses_and_save(expenses: List[Dict], task_log: TaskLog, workspace:
         expenses_object_ids = [expense_object.id for expense_object in expense_objects]
         final_query = construct_expense_filter_query(expense_filters)
 
-        mark_expenses_as_skipped(final_query, expenses_object_ids, workspace)
-        async_post_accounting_export_summary(workspace.fyle_org_id, workspace.id)
+        skipped_expenses = mark_expenses_as_skipped(final_query, expenses_object_ids, workspace)
```
ashwin1111, line 62: skipped_expense_ids

apps/quickbooks_online/exceptions.py
```diff
@@ -157,6 +158,7 @@ def new_fn(*args):
                 if not bill_payment:
                     update_failed_expenses(expense_group.expenses.all(), True)
 
+            post_accounting_export_summary(expense_group.workspace.fyle_org_id, expense_group.workspace_id, [expense.id for expense in expense_group.expenses.all()], expense_group.fund_source, True)
```
ashwin1111, line 161: which case does this handle?

apps/workspaces/actions.py
```diff
@@ -261,6 +262,7 @@ def export_to_qbo(workspace_id, export_mode=None, expense_group_ids=[], is_direc
             for expense_group_id in expense_group_ids:
                 expense_group = ExpenseGroup.objects.get(id=expense_group_id)
                 update_failed_expenses(expense_group.expenses.all(), False)
+                async_post_accounting_export_summary(expense_group.workspace.fyle_org_id, workspace_id)
```
ashwin1111, line 265: why should this be async and why inside a for loop?

ruuushhh: this was causing circular import error, also after this the function return so this being asyn call won't cause any damage.

inside for loop because we are sending based on expense group and for loop is going through expense group 
If we go out side expense group then we wil have to iterate through all expense groups to get all expense ids and then call the function.

apps/quickbooks_online/exceptions.py
ruuushhh: this is to handle all the FAILED cases, we go handle qbo exception in case of all the failures be it mapping error or accounting error, this function call will take care of posting for the errors

apps/fyle/tasks.py
ruuushhh: will make this change

ashwin1111 (approved):

PR #742, opened 2025-02-10
fix: update flag in qbo_credentials and integrations on token expiry

JustARatherRidiculouslyLongUsername:
### Description
Everytime a qbo refresh token expires, make sure that
1. `qbo_credentials.is_expired` is set to True
2. The `integrations` record is patched

## Clickup
https://app.clickup.com/t/86cxhfy2k

## Summary by CodeRabbit

- **Refactor**
  - Centralized and streamlined error management and credentials invalidation for integrations.
  - Updated authentication flows to eliminate redundant handling of token updates.
  
- **New Features**
  - Introduced a unified approach to update integration settings during error conditions.
  
- **Tests**
  - Expanded test coverage to validate the improved handling of credentials and integration settings.

### DIFF
--- a/apps/exceptions.py
+++ b/apps/exceptions.py
@@ -2,7 +2,6 @@ import logging
 
 from fyle.platform.exceptions import InvalidTokenError as FyleInvalidTokenError
 from fyle.platform.exceptions import NoPrivilegeError
-from fyle_qbo_api.utils import invalidate_qbo_credentials
 from qbosdk.exceptions import InvalidTokenError, WrongParamsError
 from rest_framework.response import Response
 from rest_framework.views import status
@@ -35,7 +34,6 @@ def handle_view_exceptions():
 
             except WrongParamsError as exception:
                 logger.info('QBO token expired workspace_id - %s %s', kwargs['workspace_id'], {'error': exception.response})
-                invalidate_qbo_credentials(kwargs['workspace_id'])
                 return Response(data={'message': 'QBO token expired workspace_id'}, status=status.HTTP_400_BAD_REQUEST)
 
             except NoPrivilegeError as exception:
@@ -44,7 +42,6 @@ def handle_view_exceptions():
 
             except InvalidTokenError as exception:
                 logger.info('QBO token expired workspace_id - %s %s', kwargs['workspace_id'], {'error': exception.response})
-                invalidate_qbo_credentials(kwargs['workspace_id'])
                 return Response(data={'message': 'QBO token expired workspace_id'}, status=status.HTTP_400_BAD_REQUEST)
 
             except Workspace.DoesNotExist:
--- a/apps/exceptions.py
+++ b/apps/exceptions.py
@@ -2,7 +2,7 @@ import logging
 
 from fyle.platform.exceptions import InvalidTokenError as FyleInvalidTokenError
 from fyle.platform.exceptions import NoPrivilegeError
-from apps.workspaces.actions import patch_integration_settings
+from fyle_qbo_api.utils import invalidate_qbo_credentials
 from qbosdk.exceptions import InvalidTokenError, WrongParamsError
 from rest_framework.response import Response
 from rest_framework.views import status
@@ -35,7 +35,7 @@ def handle_view_exceptions():
 
             except WrongParamsError as exception:
                 logger.info('QBO token expired workspace_id - %s %s', kwargs['workspace_id'], {'error': exception.response})
-                invalidate_token(kwargs['workspace_id'])
+                invalidate_qbo_credentials(kwargs['workspace_id'])
                 return Response(data={'message': 'QBO token expired workspace_id'}, status=status.HTTP_400_BAD_REQUEST)
 
             except NoPrivilegeError as exception:
@@ -44,7 +44,7 @@ def handle_view_exceptions():
 
             except InvalidTokenError as exception:
                 logger.info('QBO token expired workspace_id - %s %s', kwargs['workspace_id'], {'error': exception.response})
-                invalidate_token(kwargs['workspace_id'])
+                invalidate_qbo_credentials(kwargs['workspace_id'])
                 return Response(data={'message': 'QBO token expired workspace_id'}, status=status.HTTP_400_BAD_REQUEST)
 
             except Workspace.DoesNotExist:
@@ -69,17 +69,3 @@ def handle_view_exceptions():
     return decorator
 
 
-def invalidate_token(workspace_id, qbo_credentials=None):
-    try:
-        if not qbo_credentials:
-            qbo_credentials = QBOCredential.get_active_qbo_credentials(workspace_id)
-
-        if qbo_credentials:
-            if not qbo_credentials.is_expired:
-                patch_integration_settings(workspace_id, is_token_expired=True)
-            qbo_credentials.refresh_token = None
-            qbo_credentials.is_expired = True
-            qbo_credentials.save()
-    except QBOCredential.DoesNotExist as error:
-        logger.error(f'QBO credentials not found for {workspace_id = }:', )
-        logger.error(error, exc_info=True)
--- a/apps/exceptions.py
+++ b/apps/exceptions.py
@@ -2,6 +2,7 @@ import logging
 
 from fyle.platform.exceptions import InvalidTokenError as FyleInvalidTokenError
 from fyle.platform.exceptions import NoPrivilegeError
+from apps.workspaces.actions import patch_integration_settings
 from qbosdk.exceptions import InvalidTokenError, WrongParamsError
 from rest_framework.response import Response
 from rest_framework.views import status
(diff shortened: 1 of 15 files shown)


apps/quickbooks_online/tasks.py
```diff
@@ -726,9 +727,21 @@ def check_qbo_object_status(workspace_id):
...
                     bill.save()
+
+    except QBOCredential.DoesNotExist as error:
+        logger.error(f'QBO credentials not found for {workspace_id = }:', )
+        logger.error(error, exc_info=True)
+
     except (WrongParamsError, InvalidTokenError) as exception:
         logger.info('QBO token expired workspace_id - %s %s', workspace_id, {'error': exception.response})
 
+        if qbo_credentials:
```
ruuushhh, line 738: why not use it here ??
`invalidate_token`

apps/quickbooks_online/tasks.py
```diff
@@ -797,9 +810,21 @@ def async_sync_accounts(workspace_id):
...
         qbo_connection.sync_accounts()
+
+    except QBOCredential.DoesNotExist as error:
+        logger.error(f'QBO credentials not found for {workspace_id = }:', )
+        logger.error(error, exc_info=True)
+
     except (WrongParamsError, InvalidTokenError) as exception:
         logger.info('QBO token expired workspace_id - %s %s', workspace_id, {'error': exception.response})
 
+        if qbo_credentials:
```
ruuushhh, line 821: same here

apps/quickbooks_online/tasks.py
JustARatherRidiculouslyLongUsername: No particular reason, just that `qbo_credentials` was already initialised above

apps/quickbooks_online/tasks.py
JustARatherRidiculouslyLongUsername: No particular reason, just that `qbo_credentials` was already initialised above

apps/quickbooks_online/tasks.py
ruuushhh: let's do in function only
If the qbo creds are initialised you can pass that as a optional param
LMK if you have any questions

apps/quickbooks_online/tasks.py
JustARatherRidiculouslyLongUsername: fixed in https://github.com/fylein/fyle-qbo-api/pull/742/commits/de50d4bd4650012b1705c1334e116dc28e482898

apps/quickbooks_online/tasks.py
JustARatherRidiculouslyLongUsername: fixed in https://github.com/fylein/fyle-qbo-api/pull/742/commits/de50d4bd4650012b1705c1334e116dc28e482898

apps/exceptions.py
```diff
@@ -64,3 +67,19 @@ def new_fn(*args, **kwargs):
         return new_fn
 
     return decorator
+
+
+def invalidate_token(workspace_id, qbo_credentials=None):
+    try:
+        if not qbo_credentials:
+            qbo_credentials = QBOCredential.get_active_qbo_credentials(workspace_id)
```
ruuushhh, line 75: this is wrong 
this will always return you the active qbo creds
please check this

ruuushhh (approved):

apps/exceptions.py
```diff
@@ -64,3 +67,19 @@ def new_fn(*args, **kwargs):
         return new_fn
 
     return decorator
+
+
+def invalidate_token(workspace_id, qbo_credentials=None):
```
ashwin1111, line 72: this func should live in other file, probably helpers

ashwin1111: invalidate_qbo_credentials()

apps/quickbooks_online/tasks.py
```diff
@@ -726,8 +727,14 @@ def check_qbo_object_status(workspace_id):
                     bill.paid_on_qbo = True
                     bill.payment_synced = True
                     bill.save()
+
+    except QBOCredential.DoesNotExist as error:
+        logger.error(f'QBO credentials not found for {workspace_id = }:', )
```
ashwin1111, line 732: is this logging format right? also, level can be info

apps/quickbooks_online/tasks.py
```diff
@@ -726,8 +727,14 @@ def check_qbo_object_status(workspace_id):
...
+        logger.error(error, exc_info=True)
```
ashwin1111, line 733: this is not needed, we know that creds doesn't exist from above logger

apps/quickbooks_online/tasks.py
```diff
@@ -726,8 +727,14 @@ def check_qbo_object_status(workspace_id):
...
+
     except (WrongParamsError, InvalidTokenError) as exception:
         logger.info('QBO token expired workspace_id - %s %s', workspace_id, {'error': exception.response})
+        invalidate_token(workspace_id, qbo_credentials)
```
ashwin1111, line 737: separate out wrong params error and invalid token error here, wrong params error can be for legit cases too

apps/quickbooks_online/tasks.py
```diff
@@ -797,8 +804,14 @@ def async_sync_accounts(workspace_id):
 
         qbo_connection = QBOConnector(credentials_object=qbo_credentials, workspace_id=workspace_id)
         qbo_connection.sync_accounts()
+
+    except QBOCredential.DoesNotExist as error:
+        logger.error(f'QBO credentials not found for {workspace_id = }:', )
+        logger.error(error, exc_info=True)
```
ashwin1111, line 810: same comments as above, also why are we not invalidating it?

apps/quickbooks_online/views.py
```diff
@@ -78,7 +79,8 @@ def post(self, request, *args, **kwargs):
 
         # Check for a valid workspace and qbo creds and respond with 400 if not found
         Workspace.objects.get(id=kwargs['workspace_id'])
-        QBOCredential.get_active_qbo_credentials(kwargs['workspace_id'])
+        quickbooks_credentials = QBOCredential.get_active_qbo_credentials(kwargs['workspace_id'])
+        QBOConnector(quickbooks_credentials, workspace_id=kwargs['workspace_id'])
```
ashwin1111, line 83: this is not needed, UI does a token health call already and that takes care of invalidating

apps/quickbooks_online/views.py
```diff
@@ -98,7 +100,8 @@ def post(self, request, *args, **kwargs):
...
```
ashwin1111, line 104: same as above

apps/quickbooks_online/views.py
JustARatherRidiculouslyLongUsername: I tested this, and we were not reaching the `except:` block of `handle_view_exceptions` when we hit this endpoint. Sync and refresh dimensions were silently failing in case there was an invalid token error

I think `workspaces/:id/qbo/preferences/` is the token health check endpoint. If that is sufficient, should we remove `invalidate_token` from `@handle_view_exceptions`?

apps/quickbooks_online/tasks.py
ashwin1111: wrong params also comes for legit cases, so it's a risk to add here. There are multiple ways token can be revoked, so let's just leave this place empty for now and move out invalid token error and invalidate there

apps/quickbooks_online/views.py
ashwin1111: yes correct, `qbo/preferences/` should do, green to remove invalidate from view decorator

apps/exceptions.py
JustARatherRidiculouslyLongUsername: Fixed in https://github.com/fylein/fyle-qbo-api/pull/742/commits/598049932df23b494dfa8757e09649875dd0692e

apps/quickbooks_online/tasks.py
JustARatherRidiculouslyLongUsername: fixed in https://github.com/fylein/fyle-qbo-api/pull/742/commits/6c674a878991fcc1f1c3395cf11e54360a98bd24

apps/quickbooks_online/views.py
JustARatherRidiculouslyLongUsername: removed in https://github.com/fylein/fyle-qbo-api/pull/742/commits/a1f6ef79580196c1659d856114ad681ac1870fce

apps/quickbooks_online/tasks.py
JustARatherRidiculouslyLongUsername: fixed in https://github.com/fylein/fyle-qbo-api/pull/742/commits/5aa4246ba3b759cb98f4157f1140d872447e0a31

apps/quickbooks_online/views.py
JustARatherRidiculouslyLongUsername: removed in https://github.com/fylein/fyle-qbo-api/commit/a1f6ef79580196c1659d856114ad681ac1870fce

ashwin1111 (approved):

